Issue #98, opened 2014-03-16
Trimming is too slow


sirian:
`->trim()` method is too slow

PR #111, opened 2014-04-03
Added support for customizable stepSizes

holtkamp:
As suggested in https://github.com/Intervention/image/issues/98


holtkamp:
one year ping, I know this PR is closed, but are there viable ways to deal with this new feature? We use it in production on a daily basis for a year already and would like to not maintain a fork just for this ;).

PR #186, opened 2014-07-22
Add line spacing

lonnylot:
This is similar to the `line-height` property in CSS. It is especially useful when trying to convert an HTML/CSS presentation into an image.

Note: Though line spacing is similar to CSS `line-height` it is important to know that there are differences. CSS `line-height` is calculated as the distance from one baseline to the next baseline. Line spacing is measured as the distance from the baseline to the top of the glyph (read: character) bounding box.


wsheldon:
Was this ever added?  I didn't see it in the code.  Thanks!

PR #289, opened 2014-12-27
Addition of ability to apply stroke to drawn text

Mesuva:
Addition of functions and variables to set and draw outlines on text
(Imagick only, GD will ignore).


olivervogel:
Thank you, will be included in next release.

Issue #233, opened 2014-09-25
support webp image


thnt:
any plan to support webp image from Google?

philippzentner:
Any progress here?

razor2611:
Any news on this? We really need webp support.

garygreen:
Should be very easy to add support for webp.... Imagick support it and also [GD](http://php.net/manual/en/function.imagewebp.php)

abhimanyusharma003:
:+1:

philokezzar:
+1 for webp support please

Teranode:
anything on this?

olivervogel:
The problem is, support for WebP is only half-baked in GD and Imagick. And an implementation into Intervention Image would be only possible with hacks and workarounds.

For a better understanding I listed some of the issues. GD has `imagewebp()` and `imagecreatefromwebp()` bit apparently lacks support with other functions Intervention Image uses.

## GD issues

- `getimagesize()` does not recognize webp format
- There is no official mime type for WebP and GD function `image_type_to_mime_type()` does not recognize WebP

## Imagick issues

- Is only supported if Imagick is rebuilt with an additional webp library.

## General issues

- PHP Fileinfo extension can't read WebP inofficial mime type

paslandau:
Is there any viable way to work around this limitation? I wouldn't mind calling any native PHP functions - just wondering if anyone has already done this and would like to share :)

Cheers
Pascal

ifox:
Any update on this ?

mkwsra:
:+1:

blueset:
Looking forward for support with WebP.

abhimanyu003:
@olivervogel any plans on adding `webp` support in 3.0 at least for `imagick` users ?

hanspagel:
would love to have webp in intervention/image, even if not all features are supported yet

paulocastellano:
+1

tma-dathi:
+1

bokideckonja:
+1

bokideckonja:
Any update on this?

anik786:
+1

megalogi:
+1

d3x773r:
+1

websitevirtuoso:
+1

PR #299, opened 2015-01-10
Add webp support for GD driver

conroyp:
[`imagewebp`](http://php.net/manual/en/function.imagewebp.php) was added in php 5.5, so add webp support for the GD driver (handles half of https://github.com/Intervention/image/issues/233).

There's a bit more code here than one might initially expect, as while webp support is now technically in GD, the way it's been implemented seems to be a little inconsistent in parts. So while processing functions like `imagewebp` and `imagecreatefromwebp` exist, there is no `IMAGETYPE_WEBP` constant, so functions like `image_type_to_mime_type` and `getimagesize` don't work as smoothly as with more traditional image types. Mime type detection reports a webp image as `application/octet-stream`. So we have to override the mime type to `image/webp` when we have `application/octet-stream` and either have loaded the image so recognise the webp [magic bytes](https://developers.google.com/speed/webp/docs/riff_container#webp_file_header) at the start of the file, or fall back to file extension checking.


olivervogel:
Thank you. Will be included in the next release.

pierrejoye:
Hm, you have a good point. These constants are not all in GD and the mime functions neither.

I will see what I can do get them updated in 5.5+. Idea being to centralized all that in 7.

pfbernardo:
Hi guys,

thanks for this amazing library. Quick question:

what is the release date for the next version and will it include this code?

Thanks.

garygreen:
Any news on when PR will be merged?

garygreen:
@conroyp is it possible to support webp in ImageMagik? Should be easier than gd....?

olivervogel:
See https://github.com/Intervention/image/issues/233

PR #305, opened 2015-01-14
Display error details，"Unable to read image from path xxx" is ambiguous。

fengqi:
“no decode delegate for this image format `/path.jpg' @ error/constitute.c/ReadImage/532 (View: /path.php)”

better than 

“Unable to read image from path /path.jpg”


olivervogel:
Sorry, I disagree.

PR #335, opened 2015-03-09
lowered the resource usage of GetSizeCommand() which fixes some odd Imag...

Sm0ke0ut:
...ick segfaults


Sm0ke0ut:
Imagick tends to give segfaults with some images in combination with identifyImage(). For the specific usecase getImageWidth() and getImageHeight() should be sufficient (and is most likely less resource hunry).

This fixes a bug described on http://stackoverflow.com/questions/9873585/identifyimage-in-imagemagick-resetting-connection-in-php

olivervogel:
Makes sense. Thanks.

PR #341, opened 2015-03-16
progressive loading on .jpg

rikless:
http://php.net/manual/en/function.imageinterlace.php


olivervogel:
I don't like the idea of progressive jpeg by default.

PR #348, opened 2015-04-10
Read stream

ramiroaraujo:
Simply this, an image init option, implemented in AbstractDecoder, to read an image stream. I've implemented the same kind of tests set for other types of images


olivervogel:
Thanks.

PR #351, opened 2015-04-14
Return array of iptc keywords instead of just the first

chrisnharvey:
When getting the iptc data from an image, only one "Keyword" is returned. This PR allows multiple keywords to be returned in an array.


olivervogel:
Thanks.

PR #366, opened 2015-05-15
Face Detection Feature Added

softon:
Added face detection feature for images.

<code>$image->face()</code>

<code>$image->face(20,40)</code>

face(pad_width,pad_height)


olivervogel:
Thanks for your effort. It's nice to see that face detection is possible with GD and Imagick.

However I think face detection is more like a like a special interest feature and deserves a dedicated package, which can be installed, if the feature is needed.

Maybe you can port your code to create a separate package.

Thanks again.

softon:
Ok that seems to be good idea too.

Thx.

PR #372, opened 2015-05-22
Added being cached informations

codel:
Allows people to know if an image has already been cached or is being cached (works along with ImageCache to set the $cached property). That way they can manage the headers (for example) being sent to the browser in case the image already exists in cache.

### DIFF
--- a/src/Intervention/Image/Image.php
+++ b/src/Intervention/Image/Image.php
@@ -48,6 +48,15 @@ namespace Intervention\Image;
  */
 class Image extends File
 {
+
+    /**
+     * Indicates if the image has already been cached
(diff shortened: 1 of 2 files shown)


olivervogel:
Thank you for your PR. 

I [just added](https://github.com/Intervention/imagecache/commit/27d2ee028672c13bedf896b4ca456b7c89d84086) something like this in another way.

If an image object comes **from the cache** it is now an instance of:

`Intervention\Image\CachedImage`

If the image object has been **processed by GD or Imagick** it is.

`Intervention\Image\Image`

This way you can see, if an image was already cached.

Issue #353, opened 2015-04-20
Problem with imagick and resize


peters59:
if I resize with  imagick the images get an stairstepping effect
with gdlib it works!

see images below

original image:
![original_2015-04-17_164042](https://cloud.githubusercontent.com/assets/8724696/7225503/94e4a3a8-e73f-11e4-8ba7-d6c6ad1a8a10.png)

resize with imagickt
![imagik_2015-04-17_164112](https://cloud.githubusercontent.com/assets/8724696/7225514/a7f25daa-e73f-11e4-9882-0b139434ae26.png)

with gdlib
![gd_2015-04-17_164219](https://cloud.githubusercontent.com/assets/8724696/7225521/b54e6584-e73f-11e4-8e75-c36fb84dccf2.png)

olivervogel:
This is caused by the Imagick resize-filtering [Imagick::FILTER_BOX](https://github.com/Intervention/image/blob/master/src/Intervention/Image/Imagick/Commands/ResizeCommand.php#L23). Would be no problem to switch the filter, but I have to determine which one is efficient enough to be default and would not cause any compatibilty problems.

peters59:
Almost all filters are better than "FILTER_BOX". I think that the "FILTER_CUBIC" works well

olivervogel:
I just switched to `scaleImage` in 1e7b284f5eed963f1c6e9810d62ee3d3409cb624, which should resize without filtering, just like GD.

peters59:
"resize" again!
it would be great, if it were possible to customize the resize methode. 
at http://www.4p8.com/eric.brasseur/gamma.html#ImageMagick are many hints, how to better resize images.

see picture below, which I have resized with the default resize methode:
![sucks](https://cloud.githubusercontent.com/assets/8724696/8210178/11e8af30-1511-11e5-9f05-4f0ddc45a176.png)


BtW the fit methode still use FILTER_BOX

PR #378, opened 2015-06-09
Change Imagick method for fitCommand

aotd1:
Same fix as in #353, Fit command used resize with FILTER_BOX filter instead of scale.


olivervogel:
Thanks.

PR #390, opened 2015-06-28
Psr7 stream and response

wolfy-j:
Hi, I’v added the ability to generate [PSR-7] (http://www.php-fig.org/psr/psr-7/) compatible Stream and Response using image encoded data (methods **stream()** and **psrReponse()** accordingly). PSR7 support seems to be much more universal than laravel compatible "response" method.

I'm using [Guzzle implementation] (https://packagist.org/packages/guzzlehttp/psr7) as the easiest available option, howewer there is no PSR7 implementation exist with PHP 5.3 support.

Minimal requirements start with 5.4 so I had to change the required php version. Since 5.3 has not been supported for almost a year, it doesn't seem to be big loss for me, but it's up to author to decide.


olivervogel:
Thank you. One question before I pull this in. Why is `guzzlehttp/psr7` in `require-dev`, I should not use `stream()` when I installed with `--no-dev`?

wolfy-j:
Oh, i just copied from my repo composer.json and inserter into wrong place. :/ Fixed in latest commit.

Sorry about that.

olivervogel:
Thanks again.

PR #393, opened 2015-07-09
Add Lumen ServiceProvider

barryvdh:
Removes config_path()/publishes calls


barryvdh:
Or we could wrap those calls in `function_exists('config_path')`

olivervogel:
Thanks I removed the imagecache bootstrap stuff from the `ServiceProviderLumen`. And I added Lumen support in the the default ServiceProvider, so you can register without worrying about the name.

```php
$app->register(Intervention\Image\ImageServiceProvider::class);
```

PR #411, opened 2015-08-15
Fix isEncoded() Return Value

diarmuidie:
`isEncoded()` always returns true, even if no encoded image exists.

The `encoded` property of `Intervention\Image\Image` is initialised as an empty string:
```
    /**
     * Last image encoding result
     *
     * @var string
     */
    public $encoded = '';
```
But `isEncoded()` checks if it is null:
```
    /**
     * Checks if current image is already encoded
     *
     * @return boolean
     */
    public function isEncoded()
    {
        return ! is_null($this->encoded);
    }
```

Change the function to check for an empty string instead. Added unit test coverage.

### DIFF
--- a/src/Intervention/Image/Image.php
+++ b/src/Intervention/Image/Image.php
@@ -268,7 +268,7 @@ class Image extends File
      */
     public function isEncoded()
     {
-        return ! is_null($this->encoded);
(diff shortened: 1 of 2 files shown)


olivervogel:
Thanks.

PR #432, opened 2015-09-29
Add encode as string support

kimausloos:
Hi! I added support to encode the image as a regular string. Please let me know if there are issues with it.

If you are OK to merge this, I'll open a PR for the docs repo to add the option there too.


olivervogel:
Sorry, I think this method is obsolete, because you can always cast an image object to string.

```php
// to string
$string = (string) $image;

// or
$string = (string) $image->encode('png');
```

PR #436, opened 2015-10-05
Forcing format on save

piokom123:
Little modification that will allow us to specify format on \Intervention\Image\Image::save() like it's possible with \Intervention\Image\Image::encode()

### DIFF
--- a/src/Intervention/Image/Image.php
+++ b/src/Intervention/Image/Image.php
@@ -124,9 +124,10 @@ class Image extends File
      *
      * @param  string  $path
      * @param  integer $quality
+     * @param  string  $format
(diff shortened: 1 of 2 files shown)


olivervogel:
You force the format, by providing an corresponding filename.

```php
// save as jpeg
$image->save('foo.jpg');

// save as png
$image->save('foo.png');
```

piokom123:
In some cases it's not possible. Eg. in my current project files are saved in somename.data format.

olivervogel:
I suggest, to use `encode()` to get the image data and save it manually (file_put_contents or Laravel's great Filesystem) or even better save with an appropriate file extension.

PR #480, opened 2015-12-21
Add .gitattributes file

vinkla:
This will exclude files we don't need in production like the `tests` folder for example.


localheinz:
Blocked by #491.

olivervogel:
Thanks.

vinkla:
:+1:

PR #486, opened 2015-12-29
Update README.md to include destroy()

alexwybraniec:
Worth pointing out that allocated resources are only freed when destroy() is called, and not doing so can lead to memory leaks.


olivervogel:
Sorry, destroy() is explained in the docs. The readme is just for a quick summary.

PR #488, opened 2016-01-02
Fix: Case-insensitive method calls and class names

localheinz:
:exclamation: Blocked by #491.

This PR

* [x] fixes a case-insensitive method call
* [x] fixes case-insensitive class names


olivervogel:
Thanks.

PR #490, opened 2016-01-02
Fix: No need for leading slash in imports

localheinz:
This PR

* [x] removes leading slashes from imports

:information_desk_person: They aren't needed!


localheinz:
@olivervogel 

Rebased!

PR #491, opened 2016-01-02
Fix: Enable imagick extension on PHP5.4

localheinz:
This PR

* [x] enables the `imagick` extension on PHP5.4 after installing it

:information_desk_person: Apparently the builds currently fail because installing the extension does not automatically enable it on PHP5.4.

For reference, see 

* https://docs.travis-ci.com/user/languages/php/#Preinstalled-PHP-extensions
* http://stackoverflow.com/a/28686068/1172545


olivervogel:
Thanks.

PR #492, opened 2016-01-02
fixed Image::encode() leaks memory

olivervogel:


tinpont:
This may affect who wants save multi size of same image.

olivervogel:
Can you explain? Maybe with a code example.

tinpont:
Before this commit, I could save avatar like this:
```php
     foreach ([128, 64] as $size) {
            $avatarFile = str_replace('{size}', $size, $pathName);
            $image->widen($size)->save($avatarPath . $avatarFile);
    }
```

After this commit, maybe I just can:
```php
   $image->backup();
   foreach ([128, 64] as $size) {
       $image->reset();

       $avatarFile = str_replace('{size}', $size, $pathName);
       $content = $image->widen($size)->save($avatarPath . $avatarFile);
   }
   $image->destroy();
```

olivervogel:
Thanks, I look into it.

PR #496, opened 2016-01-09
Chain Imagick exceptions instead of hiding

goto-bus-stop:
Hi!  :v: Throwing an Exception from a `catch` block hides the original exception message. Sometimes Imagick exceptions contain useful information, like if your installation isn't compiled with PNG support. Changing this to point to the original exception using PHP's Exception chaining could make such problems more clear.

This is the only place where a different exception is thrown from a `catch` block, so I didn't touch anything else :)


olivervogel:
You're right. Thanks.

PR #510, opened 2016-02-01
Allow configure driver name as either string or array.

ajiwo:
```php

Image::configure(array('driver' => 'imagick'));

// Allow this form
Image::configure(array('driver' => array('imagick')));

// Three drivers
Image::configure(array('driver' => ['gd', 'imagick', 'dummy']));

```

### DIFF
--- a/src/Intervention/Image/ImageManager.php
+++ b/src/Intervention/Image/ImageManager.php
@@ -47,7 +47,16 @@ class ImageManager
      */
     public function make($data)
     {
-        return $this->createDriver()->init($data);
+        $image = null;
+        foreach ($this->createDriver() as $driver) {
+            if($driver instanceof AbstractDriver) {
+                $image = $driver->init($data);
+            }
+            if($image !== null) {
+                break;
+            }
+        }
+        return $image;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


olivervogel:
Why do you want to configure driver name as array?

ajiwo:
First, it's an influence from some arbitrary library, and a temptation to just pass something similar in pattern:

```
// seen as either declaration or invocation
func(act, opt);
func(opt);
func('literal');
```

I saw those forms, maybe from some Javascript or PHP framework, I can't recall.  

Seeing an example of the form:

`func(array('opt' => 'value'));`

I thought 'value' was a member of a collection, and was tempted to call it as:

`func(array('opt' => array('value', 'value2')));`

And the library would do the magic. My bad that i didn't read the documentation, but it's purely the effect of "Love at first sight" :)  

Second, a chained effect of "Love at first sight".
I have an app hosted on some scattered free web hosting to fetch image from a remote CCTV, which I scaled down for a preview. It randomly uses gd and imagick to scale the image down, because both gd or imagick sometime failed to read the image, or imagick/gd extension is not installed.
I use Intervention/Image to replace the existing image scaling code to make it more readable and maintainable.

Third, the less subjective reason: I want to provide a fallback mechanism for a given list of driver.
Pass `array('driver' => array('gd', 'imagick'))` to `configure` method and let the library pick whichever available on the host system.
Might be useful for a host system you don't have a control to install/enable PHP extension or unable to dynamically load an extension at runtime.

olivervogel:
Sorry I don't think this is a good idea.

PR #514, opened 2016-02-23
replace get_class with instanceof for $this->app

GuusDeGraeve:
When using get_class, in Lumen the $this->app could only be an instance of the \Laravel\Lumen\Application and not an instance of a class extending from \Laravel\Lumen\Application. When using the instanceof check people can use a custom application class extending from the main Lumen Application class. To make this work i had to remove the use statements at the top and replace the references in getProvider with the full paths.


olivervogel:
Thanks.

PR #516, opened 2016-02-24
Changed behaviour of 'isFilePath'

tuupke:
When Imagick is installed and used as the driver it can work with a
pageselect in the imagepath, note that this only works if ghostscript
is installed. But it then allows the library to quickly load a single
page instead of having to load the entire document; therefor
increasing the parsing speed of the document quite dramatically. When
ghostscript is not installed it will not break when imagick is used.

We also added another testcase to make sure this works. To get this
to work we had to mock the 'Config' facade, which has been done using
a mock proxy.

### DIFF
--- a/src/Intervention/Image/AbstractDecoder.php
+++ b/src/Intervention/Image/AbstractDecoder.php
@@ -2,6 +2,8 @@
 
 namespace Intervention\Image;
 
+use Config;
+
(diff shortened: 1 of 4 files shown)


olivervogel:
Sorry, can't merge this. Config is part of Laravel? Intervention Image should not be tied to any framework.

PR #518, opened 2016-02-25
Align the behaviour of Imagick crop command with GD

maaarghk:
The GD crop command creates a new canvas for when cropping. The upshot is that if you pass the start position -x, the image starts at x on the new canvas. ImageMagick does not have this behaviour in crop, which starts strictly at 0,0. extentImage however will duplicate the behaviour of the crop command in the GD driver, meaning both now perform exactly the same with negative input values.


olivervogel:
For some reason `extentImage` does not keep the image transparency. That is why the tests are failing, can't merge it this way.

maaarghk:
I'll take a look to see if there's another way to achieve this bugfix without using extentImage. Looks like GD makes a whole new canvas when cropping so can probably go down that road.

PR #519, opened 2016-02-25
Use static in drivers newImage()

vlakoff:
Helpful when using a custom driver which extends a base driver.


olivervogel:
Thanks.

PR #537, opened 2016-04-18
Added constants for ImageManager config

A1essandro:
Hi.

When I first tried to use the library, firstly, I wanted to find constants for configuration `ImageManager`. I further wrote `new ImageManager(['driver' => 'GD'])` and got an error.

I suggest: 
`new ImageManager([ImageManager::CONFIG_KEY_DRIVER => ImageManager::CONFIG_DRIVER_GD])`
Yes, it's long, but it is difficult to make a mistake.

### DIFF
--- a/src/Intervention/Image/ImageManager.php
+++ b/src/Intervention/Image/ImageManager.php
@@ -6,13 +6,29 @@ use Closure;
 
 class ImageManager
 {
+
+    /**
+     * Config key 'driver'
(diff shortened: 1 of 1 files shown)


olivervogel:
Sorry, I think this overcomplicates things.

holtkamp:
I think it makes it more robust. I had the exact same experience last week...

PR #540, opened 2016-04-26
GD: throw NotReadableException if try read partially saved image

zobzn:
getimagesize returns correct size, but image infact is not readable


olivervogel:
Thanks alot.

tugis:
"Memory exhausted"  errors are incorrectly suppress with this pull request code.

In case there's a memory exhausted error, the execution will fail and core won't have a value assigned.

PR #583, opened 2016-07-09
Add CompareCommand

andheiberg:
Simply wanted to add http://www.imagemagick.org/script/compare.php really. But added a pretty similar although reduced implementation for GD as well.

Can't really get the Imagick test to be reasonable without instantiating a real Image class which I see you don't like. https://www.dropbox.com/s/26lyvh22im6mmq1/Screenshot%202016-07-10%2000.27.24.png?dl=1 is currently the state of it.

Any suggestions for how to test this?


olivervogel:
Thank you very much. I have noticed the issue, that with GD the results differ compared to the results of Imagick.

For example I compared the following two images:

## Image1
pop1

## Image 2
pop2

And got the following different results with GD and Imagick:

## Difference result with Imagick
imagick

## Difference result with GD
gd

While the Imagick result looks right, GD seems to perceive the JPEG artifacts as difference by mistake. Is there any way to fix this?

AndreasHeiberg:
@olivervogel I mean there is a way, but it's a bit more involved, honestly I didn't even test the GD solution as it's not what I wrote it for.

So basically what you see there is indeed JPEG artifacts. The imagick solution get's around this by applying fuzz 20%. We could similarly add similar functionality in my GD source code, but it's slightly more involved. I couldn't tell you the solution right off the bat.

src/Intervention/Image/Gd/Commands/CompareCommand.php
```diff
@@ -0,0 +1,70 @@
...
+        for ($x = 0; $x < $width; $x++) {
+            for ($y = 0; $y < $height; $y++) {
+
+                $rgb1 = imagecolorat($imageCore, $x, $y);
+                $pix1 = imagecolorsforindex($imageCore, $rgb1);
+
+                $rgb2 = imagecolorat($otherImageCore, $x, $y);
+                $pix2 = imagecolorsforindex($otherImageCore, $rgb2);
+
+                if ($pix1 !== $pix2) {
```
AndreasHeiberg: At this point we would need to come up with an algorithm saying something to the effect of. These colours are minimum 20% the same.

AndreasHeiberg:
I've added the concept of fuzz or threshold as I've now coined it to the Gd version. I've also updated the tests and added compare() to the doc block for Image.

Comparison done with Gd:

compare_result

Gd and Imagick comparisons are far from identical, but they approximate each other.

AndreasHeiberg:
How do you feel about getting this merged?

AndreasHeiberg:
Any updates?

dinbrca:
Hey Oliver,
Can you please merge this? I would also like the compare command to be in this project.
Thanks!

staticdreams:
Any updates on this merge?

andheiberg:
Any chance I could get this into master?

luca-mayven:
+1

PR #606, opened 2016-09-01
Use \Illuminate\Support\Facades\Response instead of the \Response alias

nhowell:
Using the full name avoids problems where the user has commented out the Laravel aliases or has aliased "Response" to something else.

I, for one, don't use Laravel's aliases and have commented them out. But this library currently expects the alias to be there.


olivervogel:
Thanks.

nhowell:
You're welcome. Thanks for promptly merging the PR!

Do you think you could tag a new release? I would appreciate that greatly.

PR #613, opened 2016-09-19
Missing set minetype on Imagick

tchiotludo:
Mine type is not set on Imagick during convertion like it's done on GD.
So if you convert an image from png to jpg for example, the wrong minetype is returned


tchiotludo:
:up:

tchiotludo:
Any update ?

Ryvix:
I'd just like to point out that there is also webp now in the Encoder.php which is missing from here.

Also it's trying to get the mime type of an image here but I'm not sure if that's necessary and doesn't seem to be as efficient and may run out of memory on very large images. https://github.com/Intervention/image/blob/master/src/Intervention/Image/File.php#L56

PR #614, opened 2016-09-19
Support for svg

tchiotludo:
Works well on conversion (from svg to png for example).
Seems to works on resize also from svg to svg (image is working on my desktop but chrome send me an error on the svg that I don't understand) but since imagick only embed a png version in an svh image, it seems useless to do svg to svg, it's preferable to make a conversion


DracotMolver:
What is the error that chrome return?.

tchiotludo:
Here is the svg output : 
```xml
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 20010904//EN"
  "http://www.w3.org/TR/2001/REC-SVG-20010904/DTD/svg10.dtd">
<svg width="100" height="26">
  <image id="image0" width="100" height="26" x="0" y="0"
    xlink:href="data:image/png;base64,iVBOR(more more more ...)=" />
</svg>
```
And the error from chrome : 
```
This page contains the following errors:

error on line 120 at column 7: Namespace prefix xlink for href on image is not defined
Below is a rendering of the page up to the first error.
```

And the error from firefox : 
```
XML Parsing Error: prefix not bound to a namespace
Location: https://localhost/static/img/app/front/logo.5876f10d.w_100.svg
Line Number 5, Column 3:  <image id="image0" width="100" height="26" x="0" y="0"
--^
```
The image works fine in Gnome Image Viewer.
I'm really not an export of svg so I can't say why it don't work on browsers but since the conversion svg to svg offered from Imagick is simply a svg containing a png, it' don't make really any sense to output it to svg.

DracotMolver:
For what I see there are some issues.
Rules:

- If you do not set the x or y attributes, they will be set to 0.
- If you do not set the height or width attributes, they will be set to 0.
- Having a height or width attribute of 0 will disable rendering of the image.

And, if the SVG is using a prefixed attribute like **xlink:href**, it should be added into the SVG tag, he namespace of this prefixed attribute; in this casd is **xmlns:xlink= "http://www.w3.org/1999/xlink"**. Any browser works differently, so if you can see it in Gnome Image Viewer, might be because doesn't follow thos rules in a strictly way (just a theory, don't know how works the code behind Gnome app )

tchiotludo:
it's the output done directly by imagick as I know, so imagick failed to render svg.
It seems to be the same with the command line convert.
I can throw an exception when converting to svg on intervention/image.
But it's still usefull to convert in other format (email for example that is not compatible with svg for the most of them)

tchiotludo:
:up:

tchiotludo:
Any update ?

atsimerman:
Any update ?

tchiotludo:
I think the project is dead :disappointed:

PR #654, opened 2017-01-05
Fixed issue when reading image from stream if stream is not seekable

briedis:
I tried to read image from a S3 Amazon stream, but it failed when trying to call rewind(), because that stream is not seekable.

Fix is inspired by this commit, they had the same issue: https://github.com/thephpleague/flysystem/commit/cbd2d5a9c47496a8c2b64047ec2612c88c894022


briedis:
@olivervogel are you going to tag latest PRs?

PR #659, opened 2017-01-07
add support for exif to imagick extension 

frederikbosch:
This PR adds support for `$image->exif()` when using `Imagick`, without  requiring exif extension. It uses the methods [`getImageProperty`](http://php.net/manual/en/imagick.getimageproperty.php) and [`getImageProperties`](http://php.net/manual/en/imagick.getimageproperties.php) when they are available. If the extension is available, however, it prefers the extension over the imagick methods.

### DIFF
--- a/src/Intervention/Image/Imagick/Commands/ExifCommand.php
+++ b/src/Intervention/Image/Imagick/Commands/ExifCommand.php
@@ -1,0 +1,43 @@
+<?php
+
+namespace Intervention\Image\Imagick\Commands;
+
+use Intervention\Image\Commands\ExifCommand as BaseCommand;
+
+class ExifCommand extends BaseCommand
+{
+    /**
+     * Read Exif data from the given image
+     *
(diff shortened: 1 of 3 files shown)


frederikbosch:
@olivervogel I don't know where these errors are coming from but as it seems, they were already there before my PR. I see them also in other more recent PRs. As it seems, there maybe unexpected changes in the underlying extension.

briedis:
@frederikbosch yup, that's what I thought, because my previous PR did not touch that part of the code and it started failing 🤔

frederikbosch:
@briedis However, my PHP 7 does crash when running all tests. I think there is a bug with the GD extension in one of the newer versions.

frederikbosch:
@olivervogel Thanks for merging, and creating a release. We might want to update the docs for this PR. Is there a public source for the documentation? I can PR there.

olivervogel:
You can edit the docs 

https://github.com/Intervention/docs

PR #668, opened 2017-01-18
allow intervention to use own drivers

frederikbosch:
I want to extend some of the drivers. More specifically, since exif extension does not have support for streams, I must use the Imagick driver. However, in PR #659 I choose to prefer the exif extension over Imagick to prevent any possible BC break.

So I want to implement my own driver that just extends the Imagick driver a little bit. Currently, that is not possible. With this PR, you can also pass an object to the `driver` key in the `config` array.

### DIFF
--- a/src/Intervention/Image/ImageManager.php
+++ b/src/Intervention/Image/ImageManager.php
@@ -100,15 +100,25 @@ class ImageManager
      */
     private function createDriver()
     {
-        $drivername = ucfirst($this->config['driver']);
-        $driverclass = sprintf('Intervention\\Image\\%s\\Driver', $drivername);
+        if (is_string($this->config['driver'])) {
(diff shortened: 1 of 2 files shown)


frederikbosch:
@olivervogel Do you like this solution? Maybe you want a different approach. I fixed the php 5.4 problem.

frederikbosch:
Another solution would be to pass the config to the constructor of the driver, and solve it from there.

@olivervogel Sorry to disturb you with these annoying pings. Maybe you can give a direction you'd like.

frederikbosch:
Fantastic @olivervogel thanks again! If you could create a tag.... that would make it completely awesome!

olivervogel:
Tag is coming. I still have to decide to add another new feature.

PR #672, opened 2017-01-23
Introduce base ImageException class

finwe:
Base `Exception` class allows to catch all exceptions belonging to `Intervention\Image` without unrelated `\RuntimeException`s.

Naming and namespace are a proposition.


olivervogel:
Good idea. Thanks.

PR #697, opened 2017-03-08
Fix IDE error

frederikbosch:
PHPStorm complains that mixed types cannot have a string as default value.


src/Intervention/Image/Image.php
```diff
@@ -41,7 +41,7 @@
...
- * @method \Intervention\Image\Image resizeCanvas(integer $width, integer $height, string $anchor = 'center', boolean $relative = false, mixed $bgcolor = '#000000')      Resize the boundaries of the current image to given width and height. An anchor can be defined to determine from what point of the image the resizing is going to happen. Set the mode to relative to add or subtract the given width or height to the actual image dimensions. You can also pass a background color for the emerging area of the image.
+ * @method \Intervention\Image\Image resizeCanvas(integer $width, integer $height, string $anchor = 'center', boolean $relative = false, string $bgcolor = '#000000')      Resize the boundaries of the current image to given width and height. An anchor can be defined to determine from what point of the image the resizing is going to happen. Set the mode to relative to add or subtract the given width or height to the actual image dimensions. You can also pass a background color for the emerging area of the image.
```
vlakoff: You should [amend](https://www.google.com/search?q=git+amend+commit) your commit and remove this line change.

vlakoff:
Actually the flip command accepts `2` as a parameter ([Gd](https://github.com/Intervention/image/blob/master/src/Intervention/Image/Gd/Commands/FlipCommand.php), [Imagick](https://github.com/Intervention/image/blob/master/src/Intervention/Image/Imagick/Commands/FlipCommand.php)).

Though, this value is unclear and shouldn't be promoted. At most, it should only be kept for BC.
Therefore +1 for your PR.

frederikbosch:
Thanks!

PR #709, opened 2017-04-03
[FIX] File name is longer than the maximum allowed path length in some OS

denmasyarikin:
exchange positiion checking isFilePath and isBase64 and catch the error while checking is_file


olivervogel:
Thanks.

PR #717, opened 2017-04-23
Minor phpdoc fixes

vlakoff:
As you can see in `AbstractColor->parse()`, there are many formats for specifying a background color. Case in point, you can't specify transparency with strings #424242.


src/Intervention/Image/Image.php
```diff
@@ -41,7 +41,7 @@
...
- * @method \Intervention\Image\Image resizeCanvas(integer $width, integer $height, string $anchor = 'center', boolean $relative = false, string $bgcolor = '#000000')      Resize the boundaries of the current image to given width and height. An anchor can be defined to determine from what point of the image the resizing is going to happen. Set the mode to relative to add or subtract the given width or height to the actual image dimensions. You can also pass a background color for the emerging area of the image.
+ * @method \Intervention\Image\Image resizeCanvas(integer $width, integer $height, string $anchor = 'center', boolean $relative = false, mixed $bgcolor = null)           Resize the boundaries of the current image to given width and height. An anchor can be defined to determine from what point of the image the resizing is going to happen. Set the mode to relative to add or subtract the given width or height to the actual image dimensions. You can also pass a background color for the emerging area of the image.
```
vlakoff: Using null to avoid IDE warnings. This method passes null to `AbstractColor->parse()`, that gives the real default, which is actually `#ffffff`. The [documentation](http://image.intervention.io/api/resizeCanvas) would need to be fixed too.

vlakoff:
Refs #697 and pings @frederikbosch :)

PR #719, opened 2017-04-24
Another small phpdoc fix

vlakoff:
Follow-up to #717.

Noticed this afterwards. Sorry for the verbosity, but let's not let inconsistencies stay :)


olivervogel:
Thanks.

PR #730, opened 2017-05-24
Use imagecreatefromstring when imagecreatefromjpeg fail

karim-hza:
For some image the function 'imagecreatefromstring' does'nt work beacause the image is in violation with the jpeg spec. 
https://stackoverflow.com/questions/35337709/invalid-sos-parameters-for-sequential-jpeg/43271528#43271528


olivervogel:
Thanks.

PR #732, opened 2017-06-05
update doc

crisenchou:
Sometimes,the people who use Laravel will be confused by the directory 'public',so I delete it.


olivervogel:
I don't think it is confusing.

crisenchou:
But it actually confuse some of my friends.They think the 'public' means the directory of webroot(app/public),but it means any directory,right?

PR #739, opened 2017-06-13
If no encoded, encoded it

ty666:


olivervogel:
I think I end but with wrong image data, if I encode two formats after another.

```php
// example
$img = Image::make('foo.bmp'); // initial image
$img->save('foo.jpg'); // encoded data is set to jpeg
$img->save('foo.png'); // getEncoded() will take jpeg data, even if PNG is requested
```

ty666:
sorry

PR #750, opened 2017-06-28
Get travis working

dhensby:
These changes get travis working on my fork.

~~There's a test failure, but that's out of scope for a PR just intended to get the tests running.~~

`sudo: false` - to get us on the new infrastructure
Removed apt-get installs (can't do that on `sudo: false` and they failed anyway - I'm not sure if they are really needed?
Install imagick with pecl
`composer self-update || true` stops the build failing if we can't update composer for whatever reason
`composer install`:
 - `--prefer-dist` it's faster than from source
 - `--no-progress` we don't need progress for our builds so lets turn it off for more succinct output 
 - `--no-suggest` as above, we don't need this output for builds
 - `--optimize-autoloader` to hopefully speed up builds by making the autoloader a bit faster

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: php
 
+sudo: false
+
+dist: trusty
+
 php:
   - 5.4
   - 5.5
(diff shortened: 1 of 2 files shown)


dhensby:
Edit: I've added `dist: trusty` to try to get HHVM tests running and I've bumped the `PHPUnit` version because 3.x is pretty old now

dhensby:
wow - builds are much faster on trusty and the tests pass too!

olivervogel:
Thanks alot!

olivervogel:
Thanks again @dhensby  ... I have been trying to figure this out for weeks now.

dhensby:
No problem - glad to have helped. Travis can be a pain sometimes :/

PR #757, opened 2017-07-13
Fix clones of transparent GD images with a color palette

Tioni:
Currently, cloning a transparent GD image with a color palette results in the transparency being changed to black. This happens because of an incorrect conversion to true color, while also not setting the appropriate transparency color.

With this PR, cloned GD color palette images would no longer be converted to true color and keep their transparency intact.

### DIFF
--- a/src/Intervention/Image/Gd/Driver.php
+++ b/src/Intervention/Image/Gd/Driver.php
@@ -71,12 +71,53 @@ class Driver extends \Intervention\Image\AbstractDriver
      */
     public function cloneCore($core)
     {
+        if (imageistruecolor($core)) {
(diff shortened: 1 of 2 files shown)


olivervogel:
Thank you. There seems to be an error on PHP 7.1, can you fix this before the merge?

https://travis-ci.org/Intervention/image/builds/253050144

Tioni:
Last I checked this was a php 7.1 error / bug with `imagecolorat()` returning 0 for full transparency instead of 127 or something like that. (I.e. not directly related to this pr)

Sadly I did not yet find time to investigate this further, but local testing on php 7.1 seems to confirm that this is the case.

PR #776, opened 2017-09-11
FIX PNGs are now transparent when created from clones

dhensby:
I've been having an issue with PNGs rendering with black backgrounds and narrowed it down to an issue when cloning the image resource.

```php

$manager = new ImageManager();
$image = $manager->make('/path/to/image.png');
$newImage = clone $image;
$newImage->resize($width, $height);
$newImage->save('/path/to/newimage.png');
```

In this example newimage.png would be replacing all transparent pixels with black pixels.

This is because the new canvas which is created when `Gd\Driver::cloneCore` is called doesn't set the cloned image's background to transparent and so it stays at the default (black).

This fix makes sure that all changes are applied onto a transparent background when cloning images.

### DIFF
--- a/src/Intervention/Image/Gd/Commands/BackupCommand.php
+++ b/src/Intervention/Image/Gd/Commands/BackupCommand.php
@@ -15,17 +15,8 @@ class BackupCommand extends \Intervention\Image\Commands\AbstractCommand
         $backupName = $this->argument(0)->value();
 
         // clone current image resource
-        $size = $image->getSize();
-        $clone = imagecreatetruecolor($size->width, $size->height);
-        imagealphablending($clone, false);
-        imagesavealpha($clone, true);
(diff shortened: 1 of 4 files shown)


dhensby:
I've updated my fix to copy the logic from the `BackupCommand`. Perhaps the `BackupCommand` should just call `cloneCore` or even use the `clone` keyword so that the code base stores the cloning logic in a single place.

dhensby:
Right, I've also changed the `BackupCommand`s to use native cloning so that the logic for cloning images is in a single place (keeping it DRY).

dhensby:
@olivervogel any chance of some love on this 👼

olivervogel:
Thank you very much. Just merged.

dhensby:
thanks :)

PR #779, opened 2017-09-14
add getBoxSize() for imagick

Gummibeer:
In addition to #165 it would be great if `imagick` also supports this method.


olivervogel:
Thank you.

Gummibeer:
@olivervogel No problem, can you please create a tag with this change? I need it in a project and atm I have to use my own branch.

Issue #580, opened 2016-06-29
Saving svg as transparent png


jezzdk:
Seriously, I can't wrap my head around this. Basically I can do it using only Imagick, but how can I do it with Intervention?

This code works:
```
$im = new \Imagick();
$im->setBackgroundColor(new \ImagickPixel('transparent'));
$im->readImage($source);
$im->setImageFormat("png32");
$im->resizeImage($width, $height, $im::FILTER_LANCZOS, 1);
$im->writeImage($destination);
```

With this code the background is white:
```
Image::make($source)->resize($width, $height)->save($destination);
```

I also tried this with same effect:
```
Image::canvas($width, $height)->insert(
    Image::make($source)->resize($width, $height)
)->save($destination);
```

olivervogel:
Thanks for pointing out. Is fixed in 45a41a38bd1e5290cd51ab773013e6f041b2b711.

PR #818, opened 2017-12-23
Transparent Background for Images read from Blob

chellmann:
Like in issue #580 for read from file - setting the background to transparent is needed when reading SVG from Blob to avoid a white backround


olivervogel:
Thanks

PR #819, opened 2018-01-05
bringing docblock in line with documentation, re: Image::resize()

WDD-Marv:
re: `$image->resize(300, null, function($constraint) { /* stuff */ })` and `$image->resize(300)`, this stops phpstan from flagging the null parameter examples as invalid, flagging them with `Parameter #2 $height of method Intervention\Image\Image::resize() expects int, null given.`


WDD-Marv:
thanks for merging! :D

PR #829, opened 2018-03-01
Add `null` type for w/h to resize method

alexpts:
User can pass null w/h to resize method. Examples describe in docs - http://image.intervention.io/api/resize


olivervogel:
I think it is better to go with 'mixed'

```php
resize(mixed $width, mixed $height, \Closure $callback = null) 
```

PR #833, opened 2018-03-05
Restrict the maximum rotation value to 360 degrees

lboynton:
Applying a rotation value more than this may use unnecessary server side resources. For example, rotating an image by 9999999999 degrees causes the PHP process to continue indefinitely for me.


olivervogel:
Thanks

PR #837, opened 2018-03-16
Update AbstractDriver.php

umutuluer:
fix for 'command (insert) is not available for driver (Gd)' non-english locales class name issue

### DIFF
--- a/src/Intervention/Image/AbstractDriver.php
+++ b/src/Intervention/Image/AbstractDriver.php
@@ -102,6 +102,8 @@ abstract class AbstractDriver
      */
     private function getCommandClassName($name)
     {
+        $name = mb_convert_case($name[0], MB_CASE_UPPER, 'utf-8') . mb_substr($name, 1, mb_strlen($name));
+        
         $drivername = $this->getDriverName();
         $classnameLocal = sprintf('\Intervention\Image\%s\Commands\%sCommand', $drivername, ucfirst($name));
         $classnameGlobal = sprintf('\Intervention\Image\Commands\%sCommand', ucfirst($name));


iquad:
he is right, this can fix at least turkish developers problem and it is not going to broke anything for sure...

olivervogel:
Please explain the issue, you're fixing with this PR. Thanks.

umutuluer:
This error occurred because of a letter in Turkish. First letter of $name variable should be I but it is İ so two if blocks were not executing and compiler throws exception.

@olivervogel

olivervogel:
ok Thanks.

vlakoff:
* Because of the `$name[0]`, what if the first character is multibyte?
* Why explicit UTF-8 encoding for the `mb_convert_case` but not the other functions? And since PHP 5.4 it's usually fine to just omit them.
* The `mb_strlen` is useless (and 1 character too much, but doesn't harm). Omitting it just goes to the end.

Current version:
```
mb_convert_case($name[0], MB_CASE_UPPER, 'utf-8') . mb_substr($name, 1, mb_strlen($name));
```

Fixed version:
```
mb_convert_case(mb_substr($name, 0, 1), MB_CASE_UPPER) . mb_substr($name, 1);
```
<br>

* Also, the `ucfirst`'s on lines below are now redundant.
* By the way, I certainly wouldn't recommend using class names with such characters. Open gate for so many bugs.

iquad:
@vlakoff , we ain't using class names with non-unicode characters. Problem is, in Turkish based OS, capital i is İ. And you know, your code is converting to title case, so it is being "İntervention/İmage" which is silly. 

And, your code is not working, it is still converting into i to İ.
Trust me, it is the only fix :)

vlakoff:
> we ain't using class names with non-unicode characters. Problem is, in Turkish based OS, capital i is İ

I understand now, thanks. By converting yourself to an "I", you're avoiding the OS case-insensitivity that would convert "i" to "İ". Yet, my comments on the code stay valid.

> And you know, your code is converting to title case, so it is being "İntervention/İmage" which is silly.

How that?

> And, your code is not working, it is still converting into i to İ.

The only explanation I can figure out is that `mb_convert_case('i', MB_CASE_UPPER, 'utf-8')` would give "I" whereas `mb_strtoupper('i')` would give "İ". May I ask you to try these codes? Also, may I ask you what is your defined encoding, i.e. what does `mb_internal_encoding()` give?

PR #839, opened 2018-04-08
Fixed the type hints in the phpDoc blocks

Finesse:
IDEs like PhpStorm emit warnings in some methods:

It happens because there is no type `integer` in PHP, you can't set the `integer` type to a function argument:

 
I replaced all the `integer` types with `int` so IDEs don't produce produce such warnings anymore.

### DIFF
--- a/src/Intervention/Image/AbstractColor.php
+++ b/src/Intervention/Image/AbstractColor.php
@@ -7,7 +7,7 @@ abstract class AbstractColor
     /**
      * Initiates color object from integer
      *
-     * @param  integer $value
+     * @param  int $value
      * @return \Intervention\Image\AbstractColor
      */
     abstract public function initFromInteger($value);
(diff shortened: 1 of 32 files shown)


WDD-Marv:
Seems related to #819

FinesseRus:
@WDD-Marv A similar issue but not the same

WDD-Marv:
@FinesseRus does PHPStorm produce the same warning that phpstan was producing when omitting the optional parameters ?

FinesseRus:
@WDD-Marv Yes, it does in one of your cases:

image

image

WDD-Marv:
@FinesseRus thanks for checking- I'm afraid I don't have access to PHPStorm at work :(

olivervogel:
Thanks.

Finesse:
@olivervogel Thank you for the great package and merging this PR

PR #857, opened 2018-06-18
- make private method protected

danielghitoiu:
This is useful when you want to override this method, for example when used with imagecache to set it to look for images in AWS S3 as well.


olivervogel:
Thanks.

PR #873, opened 2018-07-27
Add method to colorize to RGBA

ricardofiorani:
This PR adds the possibility to use Colorize to RGBA both on GD and Imagick.

It adds a method called `colorizeToRGBA` that accepts RGBA parameters (0-255, 0-255, 0-255, 0-100).

The difference between this and the current `colorize()` method is that the new one uses a value based on 0 to 255 and accepts also opacity.

The implementation follows both GD (http://php.net/manual/en/function.imagefilter.php) and Imagick (http://php.net/manual/en/imagick.colorizeimage.php) colorise documentation.


ricardofiorani:
The building is failing because the CI uses phpunit 4.8, however on my local machine on a fresh git clone + composer install it is installing phpunit 5.7 to which the build is passing... :(

PR #879, opened 2018-08-16
correct the ‘Content-Type’ of Response

fanybook:
when i use the `encode('png')` on .jpg, the `Content-Type` is wrong


olivervogel:
Fixed in https://github.com/Intervention/image/pull/613

Issue #882, opened 2018-08-24
Missing requirement: ext-mbstring


willemwollebrants:
I'm on a fresh vm and I didn't install the mbstring extension yet. 
```bash
Call to undefined function Intervention\Image\mb_convert_case() in [...]/vendor/intervention/image/src/Intervention/Image/AbstractDriver.php:105 
```
The dependency was introduced by PR #837 https://github.com/Intervention/image/commit/e82d274f786e3d4b866a59b173f42e716f0783eb#diff-91971e0c7772a60fab6fbe264d649d99

I think there are 3 options (all of which I'm more than willing to send a PR for :))
- add ext-mbstring as a requirement in composer.json
- check if mbstring is available before using the functions. If it's not available use standard php functions instead
- remove the code that causes the problem (= revert the PR essentially)

vlakoff:
I can't decide between these options either :)

Maybe I'd rule out option 2, as it could introduce some surprises when changing environment.

vlakoff:
I added a commit in my PR that may interest you. I'm using Symfony's [mbstring polyfill](https://github.com/symfony/polyfill-mbstring). This way we have the best of options 1 and 2 combined :)

Also, maybe the mbstring extension is not as widespread as I though, and requiring to recompile the server might irritate some people…

olivervogel:
I decided to add mbstring as dependency for the next release of Version 3.

PR #888, opened 2018-09-09
Improving the fix for support of custom commands with accentuated names

vlakoff:
Follow-up to #837, see also #882.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,8 +14,8 @@
     "require": {
         "php": ">=5.4.0",
         "ext-fileinfo": "*",
-        "ext-mbstring": "*",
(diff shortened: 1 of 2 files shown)


vlakoff:
See https://github.com/Intervention/image/pull/837#issuecomment-419685491 for the detailed explanations.

Alternatively, if you don't want to add the Symfony dependency, you may do the following:
```php
if (extension_loaded('mbstring')) {
    $name = mb_strtoupper(mb_substr($name, 0, 1)) . mb_substr($name, 1);
} else {
    $name = strtoupper(substr($name, 0, 1)) . substr($name, 1);
}
```

PR #902, opened 2018-10-09
Resize width is also nullable.

koenhoeijmakers:
The docblock of the `->resize()` method currently says that `$width` must be an `int` while it may also be `null`, this PR adds `null` to it.


olivervogel:
Thanks.

PR #903, opened 2018-10-13
initFromUrl Use Curl to download pictures first

bangbangda:
file_get_contents get an individual url request, the request takes too long. The CURL method is a good way.
show demo:
`ImageManager::make('http://thirdwx.qlogo.cn/mmopen/jJvKZ28qzw0GdxdicI2PTcp46DgfCtdhZEnMWJHJ2KmibCEKwBWkFuY1XwElNwxXQqSZu6sELdv32cKyQVeZHWKlGZX6QSZcHB/46')->resize(50, 50);`
It takes 10 seconds...

other information:
https://stackoverflow.com/questions/3629504/php-file-get-contents-very-slow-when-using-full-url

### DIFF
--- a/src/Intervention/Image/AbstractDecoder.php
+++ b/src/Intervention/Image/AbstractDecoder.php
@@ -64,25 +64,41 @@ abstract class AbstractDecoder
      */
     public function initFromUrl($url)
     {
-        
-        $options = [
-            'http' => [
-                'method'=>"GET",
-                'header'=>"Accept-language: en\r\n".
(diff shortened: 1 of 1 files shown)


bangbangda:
@olivervogel  
Hi, please confirm this PR can be merged. 
I've been troubled by this problem for a long time, Please 🥺🥺🥺

jtojnar:
Adding curl support will make it even more complicated and most people will want to use something like Guzzle anyway. I would just suggest downloading the image using your preferred method in the application and then passing the blob to `Image`.

mfn:
I agree with @jtojnar

bangbangda:
> Adding curl support will make it even more complicated and most people will want to use something like Guzzle anyway. I would just suggest downloading the image using your preferred method in the application and then passing the blob to `Image`.
Thanks, maybe this is a better way. 
PR is close.

PR #905, opened 2018-10-17
Updated save method 

LucienCorreia:
If the directory does not exist then create it

### DIFF
--- a/src/Intervention/Image/Image.php
+++ b/src/Intervention/Image/Image.php
@@ -129,8 +129,7 @@ class Image extends File
     public function save($path = null, $quality = null)
     {
         $path = is_null($path) ? $this->basePath() : $path;
-        $dir = dirname($path);
-        $dir = str_replace(strrchr($path, "/"), "", $path);
+        $dir = substr($path, 0, strrpos($path, '/'));
 
         if (is_null($path)) {
             throw new Exception\NotWritableException(
--- a/src/Intervention/Image/Image.php
+++ b/src/Intervention/Image/Image.php
@@ -129,13 +129,14 @@ class Image extends File
     public function save($path = null, $quality = null)
     {
         $path = is_null($path) ? $this->basePath() : $path;
-        $dir = substr($path, 0, strrpos($path, '/'));
 
         if (is_null($path)) {
             throw new Exception\NotWritableException(
                 "Can't write to undefined path."
             );
         }
+        
+        $dir = substr($path, 0, strrpos($path, '/'));
 
         if(!is_dir($dir)) {
             mkdir($dir, 0755, true);
--- a/src/Intervention/Image/Image.php
+++ b/src/Intervention/Image/Image.php
@@ -129,6 +129,7 @@ class Image extends File
     public function save($path = null, $quality = null)
     {
         $path = is_null($path) ? $this->basePath() : $path;
+        $dir = dirname($path);
         $dir = str_replace(strrchr($path, "/"), "", $path);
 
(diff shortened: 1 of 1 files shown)


src/Intervention/Image/Image.php
```diff
@@ -129,13 +129,18 @@ public function encode($format = null, $quality = 90)
     public function save($path = null, $quality = null)
     {
         $path = is_null($path) ? $this->basePath() : $path;
+        $dir = str_replace(strrchr($path, "/"), "", $path);
```
vlakoff: This code would replace `/foo/bar.jpg/bar.jpg` (dumb, but possible) to `/foo`

I's suggest instead `$dir = dirname($path);`

LucienCorreia: `$dir = dirname($path)` return `/foo/bar.jpg` but need to return `/foo`

vlakoff: Thanks, but you should replace your line with mine, not put both ;)

vlakoff: Another thing, it should be below the `if (is_null($path))` test ;)

LucienCorreia: Being that `dirname($path)` return `/foo/bar.jpg` but need return `/foo`, the directory without the file at the end

vlakoff: That's precisely what I intended, see:
```php
dirname('/foo/bar.jpg');         // '/foo'
dirname('/foo/bar.jpg/bar.jpg'); // '/foo/bar.jpg'
```
The 2nd example is dumb, but possible. My point was to demonstrate your code could encounter pitfalls.

vlakoff: Don't you see that in my example there is a *directory* having the same name as the file?

LucienCorreia: Why put the directory name with the same file name? No one does this!

vlakoff: That's exactly what an edge case is.

This one demonstrates your code operates fundamentally wrong, even if it gives the expected result 99% of the time. If a code *can* fail, consider it *will*. See [Murphy's law](https://en.wikipedia.org/wiki/Murphy%27s_law).

And you certainly should not disregard a potential bug because "no one does this". Many people thinking like this ended up in [the daily wtf](https://thedailywtf.com/).

LucienCorreia: Now it fails too often, in fact this is a library for developers and not for end users.

"99%" is "100%" in this case.

vlakoff: Though I'd still rather recommend `dirname()`, a proper version of your code would be:
```php
$dir = substr($path, 0, strrpos($path, '/'));
```

LucienCorreia: Yes, it gets better :)

src/Intervention/Image/Image.php
```diff
@@ -129,13 +129,18 @@ public function encode($format = null, $quality = 90)
     public function save($path = null, $quality = null)
     {
         $path = is_null($path) ? $this->basePath() : $path;
+        $dir = substr($path, 0, strrpos($path, '/'));
 
         if (is_null($path)) {
             throw new Exception\NotWritableException(
                 "Can't write to undefined path."
             );
         }
```
vlakoff: As I also said, you have to put your added line under the `if (is_null($path)) {...}` part.

That's because if `$path` is null, it would run the string functions on `null`...

PR #925, opened 2019-01-21
Add support for Symfony Response creation

TiGR:
This would make response method play nice with Symfony, Silex and any other project that uses Symfony's http stack.


olivervogel:
Thanks.

PR #952, opened 2019-06-08
Change image driver via environment

DuckThom:
- Add support to set the image driver via an environment variable `IMAGE_DRIVER`

### DIFF
--- a/src/config/config.php
+++ b/src/config/config.php
@@ -15,6 +15,6 @@ return [
     |
     */
 
-    'driver' => 'gd'
+    'driver' => env('IMAGE_DRIVER', 'gd')
 
 ];


olivervogel:
Sorry, But the `env` method is part of Illuminate/support package. I don't want to include this as a depency.

DuckThom:
Hm ok, but isn't the app already depending on `illuminate/support` by the service provider? 

`use Illuminate\Support\ServiceProvider;`

olivervogel:
Yes, but this dependency applies only to the Laravel Service Providers, which called exclusively by Laravel. The code in config.php is framework agnostic. It would be possible to include the env() call in the Service Providers though.

PR #959, opened 2019-07-03
Image resolution support (ppi).

LastDragon-ru:
This PR adds initial support to get and set the resolution of images for GD and Imagick, also it supports two resolutions units: ppi and ppcm (please see notes below).

### Methods

* `getResolution(): Resolution`
* `setResolution(\Intervention\Image\Resolution $resolution)`
* `setResolution(int $xy)`
* `setResolution(int $x, int $y)`
* `setResolution(int $xy, string $units)`
* `setResolution(int $x, int $y, string $units)`

### Notes

1. Tested on PHP 7.2, on 7.1 should work too, 7.0 (fixed and supported now) and below are officially dead and I don't see any reason to support them 😄 especially because the [`gd:imageresolution()`](https://www.php.net/imageresolution) was added in 7.2  
2. GD supports only ppi, so `setResolutions(ppcm)` will automatically convert ppcm into ppi (imagick will use original units)
2. While processing images (and even after `make()`) GD will lose original resolution 😞 technically this can be fixed too but requires a lot of work and will not be included in this PR.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         }
     ],
     "require": {
-        "php": ">=5.4.0",
+        "php": ">=7.0.0",
         "ext-fileinfo": "*",
         "guzzlehttp/psr7": "~1.1"
     },
--- a/src/Intervention/Image/Gd/Commands/GetResolutionCommand.php
+++ b/src/Intervention/Image/Gd/Commands/GetResolutionCommand.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace Intervention\Image\Gd\Commands;
+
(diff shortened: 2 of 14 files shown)


mihnsen:
Any update on this?

tp-shirtigo:
I would really appreciated if this PR can be merged. The feature would be really useful in many situations. Seems like the only issue is compatibility with PHP versions <7.0.

LastDragon-ru:
> Seems like the only issue is

this project is dead

tomvo:
Someone takes the effort to write a well-tested PR but with no response from the core team at all it's sad..

PR #974, opened 2019-09-30
laravel5 pass variable as array to ImageManager [support Laravel 6 upgrade]

pirupius:
How to replay error
- Laravel 6
- PHP 7.3

Error

    "Argument 1 passed to Intervention\Image\ImageManager::__construct() must be of the type array, null given, called in vendor/intervention/image/src/Intervention/Image/ImageServiceProviderLaravel5.php on line 51"

ImageManager.php Construct

    /**
     * Creates new instance of Image Manager
     *
     * @param array $config
     */
    public function __construct(array $config = [])
    {
        $this->checkRequirements();
        $this->configure($config);
    }

ImageServiceProviderLaravel5.php

    // create image
    $app->singleton('image', function ($app) {
         return new ImageManager($app['config']->get('image'));  // variable being passed is not an array.
    });

- Laravel version 5.7 works with the current configuration but fails with Laravel 6 when using the same PHP version of 7.3. 

- Made changes so that the config being passed to the ImageManager passes an array and tested functionality with Laravel 5.7 and Laravel 6 with PHP versions 7.1, 7.2 and 7.3

### DIFF
--- a/src/Intervention/Image/ImageServiceProviderLaravel5.php
+++ b/src/Intervention/Image/ImageServiceProviderLaravel5.php
@@ -48,7 +48,7 @@ class ImageServiceProviderLaravel5 extends ServiceProvider
 
         // create image
         $app->singleton('image', function ($app) {
-            return new ImageManager($app['config']->get('image'));
+            return new ImageManager([$app['config']->get('image')]);
         });
 
         $app->alias('image', 'Intervention\Image\ImageManager');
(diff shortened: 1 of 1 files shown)


Fossil01:
Please merge this :-)

chescos:
Hey @olivervogel - could we get this merged? Looking forward to Laravel 6 support <3

olivervogel:
Thanks for the PR. But I guess this would break backward compatibility.

I fixed the issue in 2.5.1 (abbf18d5ab8367f96b3205ca3c89fb2fa598c69e).

PR #985, opened 2019-11-07
Support JP2 images

shaneiseminger:
Sometimes JPEG images carry the image/jp2 MIME type (JPEG2000 encoding). ImageMagick supports this if it has the correct delegates compiled.


jbrown0824:
Is there a reason this isn't getting merged in?

PR #990, opened 2019-11-26
Add ext-exif in composer

jmontoyaa:
This library needs ext-exif when using php 7.3


fisharebest:
We check for the existance of `exif_read_data()` before we use it.

So this `ext-exif` shouldn't be in the `required` section, it should be in the `suggests` section.

Issue #1007, opened 2020-03-04
Able to make image from ICO file but not ICO string


ziadoz:
I've stumbled across an odd issue when making an Intervention image from a favicon ICO image.

If I make the image using the file path it works correctly:
```
$faviconFile = __DIR__ . '/bbc.ico';
file_put_contents($faviconFile, file_get_contents('https://www.bbc.co.uk/favicon.ico'));

$manager = new Intervention\Image\ImageManager(['driver' => 'imagick',]);
$manager->make($faviconFile); // Works as expected.
```

However, using a string triggers an exception:
```
$faviconString = file_get_contents('https://www.bbc.co.uk/favicon.ico');

$manager = new Intervention\Image\ImageManager(['driver' => 'imagick',]);
$manager->make($faviconString); // Doesn't work: "Unable to read image from binary data."
```

Passing a pre-configured Imagick instance from a file works too:

```
$faviconFile = __DIR__ . '/bbc.ico';
file_put_contents($faviconFile, file_get_contents('https://www.bbc.co.uk/favicon.ico'));

$imagick = new Imagick();
$imagick->readImage($faviconFile);

$manager = new Intervention\Image\ImageManager(['driver' => 'imagick',]);
$manager->make($imagick); // Works as expected.
```

However, passing a pre-configured Imagick instance from a string only works if the ICO format is specified before reading the blob:

```
$faviconString = file_get_contents('https://www.bbc.co.uk/favicon.ico');

$imagick = new Imagick();
$imagick->setFormat('ICO'); // This only works if Imagick know it's an ICO ahead of time.
$imagick->readImageBlob($faviconString);

$manager = new Intervention\Image\ImageManager(['driver' => 'imagick',]);
$manager->make($imagick); // Works as expected with format line. Without doesn't work: "ImagickException - no decode delegate for this image format `' @ error/blob.c/BlobToImage/353"
```

Could Intervention configure the format during `make()` if the file is an ICO file? That way Imagick will know it's an ICO before attempting to read the blob, allowing favicons to be loaded by string.

ziadoz:
I've closed this and opened a PR to fix the issue.

ziadoz:
I've closed this and opened a fix: https://github.com/Intervention/image/pull/1008

PR #1008, opened 2020-03-04
Support making ICO and CUR files from strings (Imagemagick)

ziadoz:
If you try to make an image from an ICO or CUR image string using the Imagemagick driver an `Unable to read image from binary data` exception is thrown. If you try from a path however it works, because Imagick can determine the image type using the file extension.

I've fixed loading from a string here by checking the first four bytes of the binary to see if the string is an ICO or CUR file, and then set the appropriate format on the Imagick object. I've added test coverage for this scenario, and also for loading ICO and CUR files using paths and Imagick objects too.

One odd thing I noticed is that `finfo_open()` returns `image/vnd.microsoft.icon` on my machine, but the Travis CI builds return `image/x-icon`. I don't know how much this matters though, as this can be dealt with in userland (there are numerous mime types for ICO, so easy enough to check).

I previously opened a bug for this, but I've closed that and opened this fix instead (there's some examples of the issue there): https://github.com/Intervention/image/issues/1007 

Thanks!


src/Intervention/Image/Imagick/Decoder.php
```diff
@@ -98,6 +105,11 @@ public function initFromBinary($binary)
         $image = $this->initFromImagick($core);
         $image->mime = finfo_buffer(finfo_open(FILEINFO_MIME_TYPE), $binary);
 
+        // set standard IANA mime
+        if ($image->mime === 'image/x-icon') {
```
ziadoz: I got different results on my machine to the Travis CI build here. The IANA standard MIME is `image/vnd.microsoft.icon`, so I've ensured that's what ends up in the `$img->mime` property if the non-standard `image/x-icon` is returned.

PR #1016, opened 2020-04-26
BUGFIX: Circle x & y arguments should be optional

mattclegg:
It appears that the arguments for $x & $y should be set to be optional as they're not required (and set to '0' by default) for both supported image libraries.

Any feedback welcome.

Ref;
- https://github.com/Intervention/image/blob/master/src/Intervention/Image/Gd/Shapes/CircleShape.php#L36
- https://github.com/Intervention/image/blob/master/src/Intervention/Image/Imagick/Shapes/CircleShape.php#L36


mattclegg:
Thanks for the feedback @olivervogel

Issue #955, opened 2019-06-21
creating my own config/image.php to use imagick actually uses GD when using Facade


glederer:
When using the Laravel integration, I create a config/image.php config file to specify that i want to use imagick and not GD. However if I use the Facade approach :

       $img = Image::make(public_path($background));

it will still use GD and not Imagick, but using dependency injection :

	public function __construct(ImageManager $imageManager)
	{
		$this->imageManager = $imageManager;
	}

     ...

     $img = $this->imageManager->make(public_path($background));

then it works fine and imagick is correctly used. There seems to be an integration problem

olivervogel:
I would use a custom service provider for this.

PR #1028, opened 2020-07-13
fix config ignoring for ImageManagerStatic #955

ahurov:


src/Intervention/Image/ImageManagerStatic.php
```diff
@@ -20,7 +20,7 @@ class ImageManagerStatic
      */
     public function __construct(ImageManager $manager = null)
     {
-        self::$manager = $manager ? $manager : new ImageManager;
+        self::$manager = $manager ? $manager : resolve(ImageManager::class);
```
mfn, line 23: You can't do this because the library is not just for Laravel, can't call a framework-specific global helper.

ahurov: ok, i'm fixed it

olivervogel:

PR #1049, opened 2020-10-13
 Bugfix (Encoding format (tmp) is not supported)

vladrub:
For Windows OS

### DIFF
--- a/src/Intervention/Image/AbstractEncoder.php
+++ b/src/Intervention/Image/AbstractEncoder.php
@@ -212,8 +212,10 @@ abstract class AbstractEncoder
      */
     protected function setFormat($format = null)
     {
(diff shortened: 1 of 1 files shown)


benjamindoe:
What's a "tmp" format in terms of images?

olivervogel:

eNzyOfficial:
Are we able to get this merged? Having the same problem going from linux to windows. Works fine on linux but getting `Encoding format (tmp) is not supported.` on windows.

eNzyOfficial:
Are we able to get this merged? Having the same problem going from linux to windows. Works fine on linux but getting `Encoding format (tmp) is not supported.` on windows. @olivervogel

PR #1051, opened 2020-10-20
Better support for Lumen: Return Illuminate\Http\Response instance

benjamindoe:
Currently, with a Lumen application, `$image->response()` returns `Symfony\Component\HttpFoundation\Response`. This change compares the `app()` container with `Illuminate\Container\Container` and correctly returns an Illuminate Response without the use of a facade.

Removing the facade use is key as Lumen applications have facades disabled by default.

### DIFF
--- a/src/Intervention/Image/Response.php
+++ b/src/Intervention/Image/Response.php
@@ -3,6 +3,7 @@
 namespace Intervention\Image;
 
 use Illuminate\Http\Response as IlluminateResponse;
+use Illuminate\Support\Facades\Response as IlluminateResponseFacade;
 use Symfony\Component\HttpFoundation\Response as SymfonyResponse;
 
 class Response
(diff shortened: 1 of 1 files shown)


mfn:
It his package officially compatible with Lumen?

The reason I ask is that the change is breaking backwards compatibility. Before this PR, you could have easily mocked (faked) the Facade for whatever reason, impossible now.

However _you can enable_ Facades in Lumen. This should be carefully considered and needs a major version bump if it goes in IMHO.

benjamindoe:
Good spot. I could easily just add it to another if statement. 

I assumed the package was compatible with Lumen due to there being a Lumen-specific service provider.

mfn:
> I assumed the package was compatible with Lumen due to there being a Lumen-specific service provider.

My apologies and I assert you're definitely right about this!

(doesn't change my technical feedback though)

benjamindoe:
The latest commit checks whether facades have been "enabled". In Lumen, calling `$app->withFacades()` essentially sets the facade app. This is already set in a full Laravel application.

mfn (commented):
I can't test this but that definitely looks better to me 😄

Issue #504, opened 2016-01-26
Scaling algorithm


atayahmet:
Hi everyone,

can i prefer algorithm when image scaling? (IMG_BICUBIC, IMG_BICUBIC_FIXED etc.)

As examples of the links below
http://php.net/manual/en/function.imagesetinterpolation.php

Thank you.

olivervogel:
You can access the GD resource with [getCore()](http://image.intervention.io/api/getCore) and apply any addition function you like.

DeepDiver1975:
This does not work because in the ResizeCommand a new image is create $modified - see https://github.com/Intervention/image/blob/8ee5f346ce8c6dcbdc7dec443486bd5f6ad924ff/src/Intervention/Image/Gd/Commands/ResizeCommand.php#L47
There is no possibility for the api caller to access the GD resource of $modified.

Mind reopening this issue @olivervogel ? THX

PR #1053, opened 2020-11-11
Add imagesetinterpolation support on GD implementation

DeepDiver1975:
refs https://github.com/Intervention/image/issues/504#issuecomment-724975605

### DIFF
--- a/src/Intervention/Image/Gd/Commands/ResizeCommand.php
+++ b/src/Intervention/Image/Gd/Commands/ResizeCommand.php
@@ -17,12 +17,13 @@ class ResizeCommand extends AbstractCommand
         $width = $this->argument(0)->value();
         $height = $this->argument(1)->value();
         $constraints = $this->argument(2)->type('closure')->value();
+        $interpolation = $this->argument(3)->type('int')->value(null);
 
         // resize box
         $resized = $image->getSize()->resize($width, $height, $constraints);
 
         // modify image
-        $this->modify($image, 0, 0, 0, 0, $resized->getWidth(), $resized->getHeight(), $image->getWidth(), $image->getHeight());
+        $this->modify($image, 0, 0, 0, 0, $resized->getWidth(), $resized->getHeight(), $image->getWidth(), $image->getHeight(), $interpolation);
 
         return true;
     }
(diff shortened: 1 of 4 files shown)


src/Intervention/Image/Gd/Commands/ResizeCommand.php
```diff
@@ -62,6 +63,10 @@ protected function modify($image, $dst_x, $dst_y, $src_x, $src_y, $dst_w, $dst_h
             imagesavealpha($modified, true);
         }
 
+        if ($interpolationAlgo && function_exists('imagesetinterpolation')) {
+            imagesetinterpolation($modified, $interpolationAlgo);
+        }
+
```
mfn, line 69: In what cases is it possible to have GD but not this method? 🤔 Just wondering, thanks

mfn: Btw, `function_exists` in other cases is actually always used to throw an exception and not silently ignore a feature, doesn't sound like a good DX to me.

mfn (commented):
I'm curious how this plays out with the `FitCommand` from `Imagick`.

Because currently they both have the same "argument signature" (number and type of args, down to default values), and this makes this uneven.

And if the Imagick FitCommand receives another argument which is a different purpose, this can be very confusing.

Just wondering 🤷‍♀️

src/Intervention/Image/Gd/Commands/ResizeCommand.php
DeepDiver1975: imagesetinterpolation has been added in php 5.5 - since this library supports lower php version ....

I'm fine to throw an exception as well ... ;-)

DeepDiver1975:
> I'm curious how this plays out with the `FitCommand` from `Imagick`.

I'll happily look into this. I first wanted to touch base with you if you agree to this contribution ;-)

mfn:
> touch base with you

Achtung: I'm not even a contributor, just a user of this lib ❗

src/Intervention/Image/Gd/Commands/ResizeCommand.php
mfn: > imagesetinterpolation has been added in php 5.5 - since this library supports lower php version ....

Ah, nice!

I think the exception definitely makes more sense then quietly ignoring it => however unlikely.

olivervogel:
Spare me your sarcasm. My spirit is that you sir can use everything I have developed as is free of charge, but you cannot expect anything more.

DeepDiver1975:
Your are right. I don't expect anything anymore.

PR #1054, opened 2020-11-11
Fixed exif data from non image path (close #343, close #745, close #764)

### LINKED ISSUE #343: Exif returns null, exif_read_data() returns not null
I'm trying to read my exif data. If I run $image->exif(), it returns null. If I run exif_read_data() on the same file, I get the full array with complete data. Any ideas?

### LINKED ISSUE #745: Unable to read exif data unless loaded from path
At the momement it's only possible to read exif data from images loaded from a path.

`exif_read_data` supports the ability to load from base64 encoded data-urls. This means it is possible, in theory, to read exif data from images that are loaded through any of the `init` methods and not just those from file paths.

Would there be any support from the maintainers of this library to add exif reading capabilities to images loaded through means other than files?

---

related: https://github.com/Intervention/image/issues/336

### LINKED ISSUE #764: Exif info ignored when loading from stream
When using Image:make() with a stream instead of a file path, exif info is ignored. 

This makes orientate() impossible, some images appear with wrong rotation.

This is an example in Laravel 5.4. 
The second example loads image with the Storage facade 

```
// this works OK
$img = Image::make($file_path_here);   
print_r($img->exif());

```

```
// this doesn't work
$img = Image::make(Storage::disk('local')->get('example.jpg'));   
print_r($img->exif());
```

edgardmessias:


Kauto:
I hope this gets merged soon

PR #1059, opened 2020-11-21
PHP 8: Add support for GdImage resource objects

Ayesh:
In [PHP 8.0, `gd` resources are changed to `\GdImage` class objects](https://php.watch/versions/8.0/gdimage). This changes the `\Intervention\Image\AbstractDecoder::isGdResource` method to accept `\GdImage` objects as well as `gd` resources.

### DIFF
--- a/src/Intervention/Image/AbstractDecoder.php
+++ b/src/Intervention/Image/AbstractDecoder.php
@@ -140,6 +140,10 @@ abstract class AbstractDecoder
             return (get_resource_type($this->data) == 'gd');
         }
 
+        if ($this->data instanceof \GdImage) {
+            return true;
+        }
+
         return false;
     }
 


mfn (approved):
I've had to do similar updates for PHP8 on code parts using GD, LGTM 👍

src/Intervention/Image/AbstractDecoder.php
```diff
@@ -140,6 +140,10 @@ public function isGdResource()
             return (get_resource_type($this->data) == 'gd');
         }
 
+        if (is_object($this->data) && $this->data instanceOf \GdImage) {
```
jtojnar, line 143: The `is_object` check is superfluous according to the [docs]
(https://www.php.net/manual/en/language.operators.type.php):

> instanceof does not throw any error if the variable being tested is not an object, it simply returns false.

I would also move this check to the top of the function as going forward, this is going to be the more common case.

Ayesh: Thanks for pointing this out. The docs are indeed correct about the current behavior, but historically, `instanceof` required an object in PHP <= 7.2. 

- https://php.watch/versions/8.0/gdimage#gdimage-is-resource
- https://3v4l.org/YlFF5

I added that test because this library supports PHP 5.4 and later.

jtojnar: That is just issue with literals, non-object variables work as far as PHP 5.0:

https://3v4l.org/32265

Ayesh: You are right, I will push with the is_object() call removed.

Ayesh:
I force-pushed an update with the `is_object()` call removed. Thanks @jtojnar for this 🙏🏼.

jtojnar:
What do you think about moving the check up before the `is_resource` check?

src/Intervention/Image/AbstractDecoder.php
```diff
@@ -140,6 +140,10 @@ public function isGdResource()
             return (get_resource_type($this->data) == 'gd');
         }
 
+        if ($this->data instanceOf \GdImage) {
```
jtojnar, line 143: Looks like the lowercase spelling of `instanceof` is preferred: https://github.com/Intervention/image/search?q=instanceof

Ayesh: This was overlooked as well, thank you. Force-pushed now.

james090500:
Do we have any idea when we can expect this to merge?

mrizkimaulidan:
merge it please for php 8 compatibility

james090500:
Any reason we aren't seeing this merged? I can't update anything to PHP 8 because of this...

Perzonallica:
Btw, this solution still doesn't seem to work. I also tried this one: https://github.com/Intervention/image/pull/1068 but I'm still having issues on PHP8:

`Exception
Backup not available. Call backup() before reset().`

Hovewer I'm using the backup() before the reset() and it works on PHP7.4:

```
        $img = Image::make(Storage::path($imageConfig['game_folder_path'] . DIRECTORY_SEPARATOR . 'big_image_original.' . $imageConfig['extension']));

        // backup status
        $img->backup();

        // Convert to jpg
        $img->save(Storage::path($imageConfig['game_folder_path'] . DIRECTORY_SEPARATOR . $imageConfig['filename'] . '.jpg'), 80, 'jpg');

        // Restore image
        $img->reset();

        // Convert to webp
        $img->save(Storage::path($imageConfig['game_folder_path'] . DIRECTORY_SEPARATOR . $imageConfig['filename'] . '.jpg.webp'), 70, 'webp');

        // Remove img from memory
        $img->destroy();
```

mtctonyhkhk2010:
Please merge this for php 8 support

pavarnos:
this was merged to master on 6 Jul, but the change is not -in- master?

PR #1062, opened 2020-12-03
Fix Decoder::initFromUrl under service mesh with envoy side raise a 426 http code

qunqiang:
my company uses this package for fasting image processing, but when we deploy to k8s with istio, the envoy sidecar requires newer HTTP version, so i think we should improve http context from HTTP 1.0 (default to HTTP Context) to HTTP 1.1 with context options 'http.protocol_version'.

### DIFF
--- a/src/Intervention/Image/AbstractDecoder.php
+++ b/src/Intervention/Image/AbstractDecoder.php
@@ -69,6 +69,7 @@ abstract class AbstractDecoder
         $options = [
             'http' => [
                 'method'=>"GET",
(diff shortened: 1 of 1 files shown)


mfn (approved):
Sounds sane to me.

Also, with PHP 8 this will actually become the default, see https://www.php.net/manual/en/migration80.incompatible.php
> The HTTP stream wrapper as used by functions like file_get_contents() now advertises HTTP/1.1 rather than HTTP/1.0 by default. This does not change the behavior of the client, but may cause servers to respond differently.

Basically exactly the reason why you made the PR 😄

PR #1063, opened 2020-12-07
Security Fix for Cross-site Scripting (XSS) - huntr.dev

huntr-helper:
https://huntr.dev/users/Asjidkalam has fixed the Cross-site Scripting (XSS) vulnerability 🔨. Think you could fix a vulnerability like this?

Get involved at https://huntr.dev/

Q | A
Version Affected | ALL
Bug Fix | YES
Original Pull Request | https://github.com/418sec/image/pull/1
Vulnerability README | https://github.com/418sec/huntr/blob/master/bounties/packagist/intervention/image/2/README.md

### 📊 Metadata *

Fixed Cross-Site scripting vulnerability

### ⚙️ Description *

`intervention/image` is an image handling and manipulation library. This package is vulnerable to XSS via reading the exif data from image.

### 💻 Technical Description *

Issue: https://github.com/Intervention/image/blob/master/src/Intervention/Image/Commands/ExifCommand.php#L31 
The issue existed because there were no checks implemented to sanitize the data received from `exif_read_data` before rendering it to the document. The fix is to sanitize them by recursively mapping the `$data` array with `htmlspecialchars()`

### 🐛 Proof of Concept (PoC) *

Use the same steps given in the bounty description, with an image updated with XSS payload on its model EXIF data.
```php
<?php
require 'vendor/autoload.php';

// import the Intervention Image Manager Class

$filename = "./photo.jpeg";

// read all existing data into an array
$data = Image::make($filename)->exif();

// read model of the camera
$name = Image::make($filename)->exif('Model');

echo $name;
```

image

### 🔥 Proof of Fix (PoF) *

After applying the committed code, it no longer triggers the XSS.

image

### 👍 User Acceptance Testing (UAT)

Just sanitized the array, no breaking changed introduced

:)

### DIFF
--- a/src/Intervention/Image/Commands/ExifCommand.php
+++ b/src/Intervention/Image/Commands/ExifCommand.php
@@ -29,7 +29,19 @@ class ExifCommand extends AbstractCommand
         // try to read exif data from image file
         try {
             $data = @exif_read_data($image->dirname . '/' . $image->basename);
-
+            
+            // recursively sanitize the data from exif
+            $data = array_map(function($element){
+                $res = null;
+                if(is_array($element)){
+                    $res = array_map('htmlspecialchars', $element);
+                }
+                else {
+                    $res = htmlspecialchars($element, ENT_QUOTES, 'UTF-8');
+                }
+                return $res;
+            }, $data);
+            
             if (!is_null($key) && is_array($data)) {
                 $data = array_key_exists($key, $data) ? $data[$key] : false;
             }


JamieSlome:
@olivervogel - let me know if you have any questions or thoughts, cheers! 🍰

mfn:
It strikes me as this fix is not correct: that particular commands concern is about "extracting data".

Sure, it might be unexpected to have HTML there, but it has to be treated as "user input", i.e. potentially malicious.

From the code sample, the fix is not to "echo" it but sanitize _there_.

Just imagine if the metadata contains an SQL injections and you don't use prepared statements, saving the cameras `Model` could wipe your database.

I know this is a clash of philosophies, but the right fix would be educated the users of this library about the fact to **not** trust embedded exit data.

IMHO

JamieSlome:
@Asjidkalam @Mik317 - it would be good to get your opinions here, thanks! 🍰

jtojnar:
Yes, this is corrupting data by assuming the output is HTML and escaping it as such – if I wanted to generate a PDF, for example, I would have to unescape the data manually. And to make it more confusing, items in `array` items are escaped differently than scalar values. And worst of all, the code assumes `UTF-8` encoding, which might not be true for EXIF data. When it is not the case, it might garble the input irreversibly.

Proper place to prevent XSS is the application (usually in the View layer) since only the application can know how are the data used.

PR #1068, opened 2021-01-07
PHP 8  \GDImage object Support

buzzclue:


mfn:
duplicate of https://github.com/Intervention/image/pull/1059

buzzclue:
@mfn That PR will still not work on PHP8, compare the changes and approach.

mfn:
Is the other one incomplete? Both PRs touch different lines… would it need both? 🤷‍♀️

buzzclue:
@mfn  \Intervention\Image\Gd\Decoder uses the AbstractDecoder abstract Class

src/Intervention/Image/Gd/Commands/ResetCommand.php
```diff
@@ -17,7 +17,7 @@ public function execute($image)
     {
         $backupName = $this->argument(0)->value();
 
-        if (is_resource($backup = $image->getBackup($backupName))) {
+        if (is_resource($backup = $image->getBackup($backupName)) || (is_object($backupName) && $backupName instanceof \GDImage)) {
```
ADmad, line 20: It would be cleaner to move the `$backup` var assignment before the `if` statement.

tms1987: According to https://php.watch/versions/8.0/gdimage it's not necessary to call is_object.
Also you should use `$image->getBackup($backupName) instanceof \GDImage`, otherwise the default reset and backup commands are not working.

olivervogel:
I merged https://github.com/Intervention/image/pull/1059, which adressed the same issue.

pavarnos:
#1059 fixed AbstractDecoder, but did not fix Gd/Commands/ResetCommand. See #1107

PR #1069, opened 2021-01-13
Security Fix for Relative Path Traversal - huntr.dev

huntr-helper:
https://huntr.dev/users/Asjidkalam has fixed the Relative Path Traversal vulnerability 🔨. Think you could fix a vulnerability like this?

Get involved at https://huntr.dev/

Q | A
Version Affected | ALL
Bug Fix | YES
Original Pull Request | https://github.com/418sec/image/pull/2
Vulnerability README | https://github.com/418sec/huntr/blob/master/bounties/packagist/intervention/image/1/README.md

### 📊 Metadata *

Fixed Directory Traversal bug.

### ⚙️ Description *

`intervention/image` is an image handling and manipulation library. This package is vulnerable to Directory Traversal attack.

### 💻 Technical Description *

A path traversal issue exists in the program which can be used to access files anywhere in the system and save them as different files. Its caused due to the improper sanitization of the filename, so the fix would be to validate the file name and check whether it contains `..` characters. 

### 🐛 Proof of Concept (PoC) *

Same as PoC given in the bounty description.

```php
<?php

require 'vendor/autoload.php';

$imageFilename = '../2/image-2.jpg';

// import the Intervention Image Manager Class

// open an image file
$img = Image::make('uploads/1/'.$imageFilename);

// now you are able to resize the instance
$img->resize(320, 240);

// finally we save the image as a new file
$img->save('uploads/1/newphoto.jpeg');
```

image

### 🔥 Proof of Fix (PoF) *

After applying the fix, it doesn't save the file anymore - instead, it throws an exception saying `Image source not readable`, which is intended.

image

### 👍 User Acceptance Testing (UAT)

Just sanitized the filename, no breaking changes introduced. 
:)

### DIFF
--- a/src/Intervention/Image/Commands/ExifCommand.php
+++ b/src/Intervention/Image/Commands/ExifCommand.php
@@ -29,7 +29,19 @@ class ExifCommand extends AbstractCommand
         // try to read exif data from image file
         try {
             $data = @exif_read_data($image->dirname . '/' . $image->basename);
-
+            
+            // recursively sanitize the data from exif
+            $data = array_map(function($element){
+                $res = null;
+                if(is_array($element)){
+                    $res = array_map('htmlspecialchars', $element);
+                }
+                else {
+                    $res = htmlspecialchars($element, ENT_QUOTES, 'UTF-8');
+                }
+                return $res;
+            }, $data);
+            
             if (!is_null($key) && is_array($data)) {
                 $data = array_key_exists($key, $data) ? $data[$key] : false;
             }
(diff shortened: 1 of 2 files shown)


mfn:
👎 

If you have an application accepting arbitrary strings without filtering and literally using them as paths, then it's the application which has a problem, but not this library.

This library is able to load images from arbitrary paths _as designed_, nothing more and nothing less. There is no vulnerability here thus there is nothing to fix.

---
This is the same with the other PR https://github.com/Intervention/image/pull/1063 . Not sure what is going on here: making up "security issues" were none are?

benjamindoe:
I feel like this guy is probably just dipping his feet into the world of bug bounties and CVEs. Typical undergrads...

https://github.com/Asjidkalam

PR #1070, opened 2021-01-15
fix: gd memory issue when retrieving mime type

chrfritsch:
I had some memory issues within the gd driver.

`Allowed memory size of 134217728 bytes exhausted (tried to allocate 10076428 bytes)`

With that little change, it works now, even with 64MB.


msEmmaMays:
This is because of a bug in Libmagic that was just fixed, it hasn't made it into mainline releases yet.

https://bugs.astron.com/view.php?id=234

```
$ php -i | grep magic

libmagic => 537

```

This needs to be 540 or higher for the fix to be in libMagic.

PR #1071, opened 2021-01-19
Fix issue where execute is failing due to is_resource no longer used …

streamingsystems:
…in php 8

I was running into a problem were I could not backup and reset an image in php 8.


streamingsystems:
Typo, sorry.

PR #1091, opened 2021-06-07
Add support to equal sign - '=' - on URL

jaircuevajunior:
Updated filename regex pattern to allow the equal sign as being part of the filename portion of the route, otherwise it would fall on 404 error.

### DIFF
--- a/src/Intervention/Image/ImageServiceProviderLaravelRecent.php
+++ b/src/Intervention/Image/ImageServiceProviderLaravelRecent.php
@@ -77,7 +77,7 @@ class ImageServiceProviderLaravelRecent extends ServiceProvider
         // imagecache route
         if (is_string(config('imagecache.route'))) {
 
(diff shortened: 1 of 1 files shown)


olivervogel:

jaircuevajunior:
Sure.
In this case I was using a base64 encoded string as filename (which sometimes ends with an equal sign).
I realized that when the equal sign is present on URL, the route is not recognized.
I'm using intervention/imagecache for this but the route is created in this package (intervention/image) service provider.

To clarify, for instance:

Let's take 'images/cache' as the imagecache route.
If I try to get images/cache/module_desk/uploads/module_avatar/KEQbRnXa47uqIFaEoUutsD98lqc5Rv-metaQmFubmVyLW1vYmlsZS5qcGc=-.jpg it doesn't work, because it fails the regex, so the route is not even called.

PR #1099, opened 2021-07-16
Fallback to Utils::streamFor for guzzlehttp/psr7 ^2.0

### LINKED ISSUE #1095: guzzlehttp/psr7 v2.0 API change breaks part of the library
Hello, 

With the new PSR7 library released a few days ago, some API has changed and it broke the StreamCommand that I was using. Is there any plan to upgrade the API in order to support the latest version of PSR7?

For your reference
https://github.com/guzzle/psr7/blob/c0dcda9f54d145bd4d062a6d15f54931a67732f9/README.md#upgrading-from-function-api

Thanks

SychO9:
**Fixes #1095**
**Fixes flarum/core#2966**

`guzzlehttp/psr7` changed the function API to a utility class `Utils`, I checked and this should be the only function from that API used in this package.

https://github.com/guzzle/psr7/blob/c0dcda9f54d145bd4d062a6d15f54931a67732f9/README.md#upgrading-from-function-api

### DIFF
--- a/src/Intervention/Image/Commands/StreamCommand.php
+++ b/src/Intervention/Image/Commands/StreamCommand.php
@@ -2,6 +2,9 @@
 
 namespace Intervention\Image\Commands;
 
+use GuzzleHttp\Psr7\Utils;
+use function GuzzleHttp\Psr7\stream_for;
+
 class StreamCommand extends AbstractCommand
 {
     /**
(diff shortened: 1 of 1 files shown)


olivervogel:
Thanks. Already fixed this issue in e2b8aebe3e5baf1e3db61a0f5227502313d74b29.

PR #1116, opened 2021-10-04
Missing 'jfif' format encoder. Only have 'image/jfif' format.

muhdfaiz:
I have an issue with JFIF format. The problem is some of the JFIF the format only return as `jfif` not `image/jfif`. Right now the code only check for `image/jfif`. I just added `jfif` in the `AbstractEncoder.php` file.


helios-ag:
@olivervogel can you please tag a new release?

PR #1133, opened 2021-12-30
Added dpi method for image conversion units

RohanSakhale:
# New Method Introduction

`dpi` helps to update image density supported only for imagick driver

It throws exception if called over other driver


RohanSakhale:
@olivervogel can you please review if this is a contribution is acceptable or requires an update

RohanSakhale:
@olivervogel been a year waiting to hear from you :-)

sumonst21 (approved):

PR #1137, opened 2022-01-29
Prevent passing of null to strtolower() : Size.php

jcogs-design:
Fix deprecation error when running under php8.1
```strtolower(): Passing null to parameter #1 ($string) of type string is deprecated
vendor/intervention/image/src/Intervention/Image/Size.php, line 281```


olivervogel:

PR #1144, opened 2022-02-24
line width support added to GD driver

llabbasmkhll:
according to [php manual](https://www.php.net/manual/en/function.imagesetthickness.php) 
line width can be set in GD driver
so i added it to line shape abstract


llabbasmkhll:
@ramiroaraujo @jwpage @mattcannon

PR #1147, opened 2022-03-26
[next] Update workflow and some improvements

kudashevs:
Hello,

I would like to propose a little contribution to the project. Sorry, for the big PR. I started it just with an intention to improve the CI/CD workflow and add PHP 8.1 support. But it turned out into a massive change. So, this PR:
- updates the ` run-tests.yml` with the possibility to run tests with the different versions of the ImageMagick library. There was a problem in the previous CI/CD because the default Ubuntu 20.04.4 ImageMagick library (version 6.9.10.23) has a bug, which hinders the tests to pass. That’s why we have to use a higher version which should be compiled from scratch (a package with a higher version doesn’t exist for Ubuntu 20.04.4, at least I didn’t find any in the official repository neither in non official repositories)
- updates `.gitignore` file with a build folder and phpunit.xml. According to my experience, it’s better to provide a `phpunit.xml.dist` file with the default settings, while the `phpunit.xml` can be used by other developers for their own purposes.
- adds the default `phpunit.xml.dist` file. If you don't like this idea, I can remove this change from the PR. Let me please know about your decision.
- updates `FitModifier.php` and `ResizeModifier.php` with the additional transparency refinements. That’s a very interesting case because the transparency after resize gives wrong color data in some cases. Therefore, the dependent tests fail. I had to investigate this situation and this is one of the explanations with [a solution](https://stackoverflow.com/questions/724503/php-resize-script-doesnt-work-for-transparent-gifs) (which I had adjusted to your code)
- fixes missing method signature in the `SizeInterface.php`
- renames `testImageDecoder` helpers

I hope this little contribution will be helpful :)


thomaspicquet (approved):

olivervogel:
Awesome. Thank you very much.

PR #1149, opened 2022-04-03
Explicitly convert to integer

nztim:
This is intended to stop deprecation notices on PHP 8.1, e.g.:
```
Implicit conversion from float 256.8493150684931 to int loses precision in /.../vendor/intervention/image/src/Intervention/Image/Gd/Commands/ResizeCommand.php on line 67 [] []
```

The other usages of these functions don't look like they will cause this problem.


olivervogel:
Thanks.

PR #1157, opened 2022-06-07
Remove a php 8.1 deprecation warning

jcogs-design:
Change implicit float->int conversion with use of round(x,0)


olivervogel:
Thanks.

PR #1163, opened 2022-07-04
Nullable type declaration for default null value

MathiasReker:
Adds ? before type declarations for parameters with a default null value


olivervogel:

PR #1165, opened 2022-07-04
Ternary to null coalescing

MathiasReker:
Use null coalescing operator ?? where possible.


olivervogel:

PR #1166, opened 2022-07-04
Visibility required

MathiasReker:
Visibility MUST be declared on all properties and methods; abstract and final MUST be declared before the visibility; static MUST be declared after the visibility.


olivervogel:

PR #1170, opened 2022-07-04
Single blank line at eof

MathiasReker:
A PHP file without end tag must always end with a single empty line feed.


olivervogel:
Thanks.

PR #1186, opened 2022-10-21
[fix] Size.php php8.1 strtolower() Passing null to param

erikn69:
I'm getting ErrorException on PHP 8.1, PHP 8.2

`strtolower(): Passing null to parameter #1 ($string) of type string is deprecated`

Without this change, the defaults are never set

https://github.com/Intervention/image/blob/54934ae8ea3661fd189437df90fb09ec3b679c74/src/Intervention/Image/Size.php#L345-L351

### DIFF
--- a/src/Intervention/Image/Gd/Font.php
+++ b/src/Intervention/Image/Gd/Font.php
@@ -229,7 +229,7 @@ class Font extends \Intervention\Image\AbstractFont
             }
 
             // x-position corrections for horizontal alignment
-            switch (strtolower($this->align)) {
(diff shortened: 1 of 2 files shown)


angeljqv:
@olivervogel  ping : )

thomaspicquet (approved):

angeljqv:
@olivervogel  ping : )

fisharebest:
> Without this change, the defaults are never set

Would it be better to set a default value on these properties?  e.g. `public $align = 'bottom';`

parallels999:
@olivervogel hi

Issue #1190, opened 2022-11-07
Orientation EXIF tag lost with Imagick and gd drivers


gmarcon:
The Intervention Image library, when initiating from a stream, loses all EXIF data with the gd driver, while it resets the Orientation EXIF tag to 0 with the Imagick driver.

The following code, used on this [test image](https://raw.githubusercontent.com/recurser/exif-orientation-examples/master/Landscape_3.jpg):
```
<?php
use Intervention\Image\ImageManager;

$driver = ['driver' => 'Imagick'];
$image_manager = new ImageManager($driver);

$image = $image_manager->make('./Landscape_3.jpg');
echo "Instantiated from file path, Imagick:\n";
echo var_export($image->exif(), TRUE);
echo "\n";

$stream = fopen('./Landscape_3.jpg', 'r');
$image = $image_manager->make($stream);
echo "Instantiated from stream, Imagick:\n";
echo var_export($image->exif(), TRUE);
echo "\n";

$driver = ['driver' => 'gd'];
$image_manager = new ImageManager($driver);

$image = $image_manager->make('./Landscape_3.jpg');
echo "Instantiated from file path, gd:\n";
echo var_export($image->exif(), TRUE);
echo "\n";

$stream = fopen('./Landscape_3.jpg', 'r');
$image = $image_manager->make($stream);
echo "Instantiated from stream, gd:\n";
echo var_export($image->exif(), TRUE);
echo "\n";
?>
```

Generates the following output:
```
Instantiated from file path, Imagick:
array (
  'FileName' => 'Landscape_3.jpg',
  'FileDateTime' => 1667820935,
  'FileSize' => 348796,
  'FileType' => 2,
  'MimeType' => 'image/jpeg',
  'SectionsFound' => 'ANY_TAG, IFD0',
  'COMPUTED' =>
  array (
    'html' => 'width="1800" height="1200"',
    'Height' => 1200,
    'Width' => 1800,
    'IsColor' => 1,
    'ByteOrderMotorola' => 1,
  ),
  'Orientation' => 3,
  'XResolution' => '72/1',
  'YResolution' => '72/1',
  'ResolutionUnit' => 2,
  'YCbCrPositioning' => 1,
)
Instantiated from stream, Imagick:
array (
  'FileDateTime' => 0,
  'FileSize' => 440610,
  'FileType' => 2,
  'MimeType' => 'image/jpeg',
  'SectionsFound' => 'ANY_TAG, IFD0',
  'COMPUTED' =>
[...]

PR #1192, opened 2022-11-07
Allow EXIF data to be read from streams.  Fixes: #1190

fisharebest:
Intervention can create images from filenames and streams (and others).

PHP:exif_read_data() can read EXIF data from both filenames and streams (since PHP 7.2).

But when intervention creates an `Image` from a stream, it reads the data (excluding the EXIF data), and then creates a new stream from just the image data.  Thus the EXIF data is lost.

We lose the original stream (and the EXIF data) here: [src/Intervention/Image/AbstractDecoder.php#L125](https://github.com/Intervention/image/blob/54934ae8ea3661fd189437df90fb09ec3b679c74/src/Intervention/Image/AbstractDecoder.php#L125)

We create a new stream from this (EXIF-less) data here: [src/Intervention/Image/Commands/StreamCommand.php#L20](https://github.com/Intervention/image/blob/54934ae8ea3661fd189437df90fb09ec3b679c74/src/Intervention/Image/Commands/StreamCommand.php#L20)

When we create an image from a file, we store the basename/dirname in the `Image` object.  (Actually, in the parent `File` object).  This original data is used to read EXIF data.

This PR uses a similar approach for images created from a `stream`.  We keep a copy of the original stream and use it when we need the EXIF data. 

I created a test case for this, using a rotated image.  This test passes.  There are a number of pre-existing test failures, which I haven't investigated.


fisharebest:
Note: I left the existing `->stream()->detach()` code in place.

I don't think it works, but it is possible it may work with custom drivers.

thomaspicquet (approved):

Samvdg:
@fisharebest Do you still use this package? I'm currently looking for an alternative because of the EXIF data issue and was wondering what you use if it's not this

fisharebest:
> Do you still use this package? I'm currently looking for an alternative because of the EXIF data issue and was wondering what you use if it's not this

I use my own fork of this package, with this fix.

Samvdg:
> > Do you still use this package? I'm currently looking for an alternative because of the EXIF data issue and was wondering what you use if it's not this
> 
> I use my own fork of this package, with this fix.

Thanks for the reply! I ended up writing my own middleware for the package. Now the issue should be fixed though

PR #1217, opened 2023-07-06
Fixed avif support

double-web:


olivervogel:
Encoding/decoding AVIF is supported in [version 3](https://image.intervention.io/v3/).

PR #1228, opened 2023-12-01
GD: add support of AVIF (read image from path)

stealeks:
Fixes

_NotReadableException
Unsupported image type image/avif. GD driver is only able to decode JPG, PNG, GIF, BMP or WebP files._

https://www.php.net/manual/en/function.imagecreatefromavif.php


olivervogel:
Encoding/decoding AVIF is supported in [version 3](https://image.intervention.io/v3/).

PR #1242, opened 2023-12-21
Accept `$quality` param in `MediaTypeEncoder`

hivokas:
Thanks for releasing v3!

I think it would be convenient to set `$quality` param for `MediaTypeEncoder`. 

If you don't like this change, it's fine, I can create this class in our project.

If you like this change, you can merge it :)

### DIFF
--- a/src/Encoders/AutoEncoder.php
+++ b/src/Encoders/AutoEncoder.php
@@ -9,6 +9,10 @@ use Intervention\Image\Interfaces\ImageInterface;
 
 class AutoEncoder implements EncoderInterface
 {
+    public function __construct(public int $quality = 75)
+    {
+    }
+
     /**
      * {@inheritdoc}
      *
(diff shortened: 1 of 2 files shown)


hivokas:
Also, is there a reason `GifEncoder` and `PngEncoder` don't support `$quality` param? It it a limitation of Imagick?

olivervogel:
> Also, is there a reason `GifEncoder` and `PngEncoder` don't support `$quality` param? Is it a limitation of Imagick?

Thanks for your PR. GIF format uses lossless compression so there is no control for quality. For the PNG format, I have decided to stick with the default value of -1 for the time being, which gives a good result for the file size. However, this could still change, but I want to investigate the results with different parameters first with both drivers. I have created  #1243  for this.

I have moved the method `encoderByMediaType()` to `MediaTypeEncoder::class` an hour ago. Please resolve this conflict and I will merge.

hivokas:
> Thanks for your PR. GIF format uses lossless compression so there is no control for quality. For the PNG format, I have decided to stick with the default value of -1 (with GD) for the time being, which gives a good result for the file size. However, this could still change, but I want to investigate the results with different parameters first with both drivers. I have created #1243 for this.

Got it, thanks for explanation.



> I have moved the method `encoderByMediaType()` to `MediaTypeEncoder::class` an hour ago. Please resolve this conflict and I will merge.

Fixed. Does it look good to you?

olivervogel:
I think we need a constructor with `quality` paramter in `AutoEncoder::class` as well. Currently you have to pass two parameters. Although the type parameter is not necessary.

```
ImageManager::gd()
    ->read('example.jpg')
    ->encode(new AutoEncoder(null, 0));
```

olivervogel:
And to make things complete, it would be great if you could also integrate the `quality` parameter in the following classes:

- `Intervention\Image\Encoders\FileExtensionEncoder::class`
- `Intervention\Image\Encoders\FilePathEncoder::class`
-  `Intervention\Image\Encoders\MediaTypeEncoder`

olivervogel:
Your suggested changes are now already integrated. See: https://github.com/Intervention/image/commit/ea13f2e0bb7c30329712d2f90a6c4dcdd12d628d

hivokas:
Thank you @olivervogel, good changes, I appreciate it.

PR #1245, opened 2023-12-30
duplicate Countable Interface in image.php

sirosfakhri:
`image.php` implemented two interface, `ImageInterface` and  `Countable`.
but `ImageInterface` interface already included `Countable`, So, this one is extra and can be deleted.


olivervogel:
You're right. Thanks.

PR #1271, opened 2024-01-16
 Implemented basic ImageMagick trim modifier

### LINKED ISSUE #1239: Trim function in v3
I understand that updates and improvements are essential, but the trim function served a crucial purpose for my projects. I kindly request the reconsideration and reimplementation of the trim function in future releases.

If there are specific reasons for its removal or alternative approaches, I would appreciate any insights you can provide. Your efforts in maintaining this library are highly appreciated, and I believe the reintroduction of the trim function would benefit many users.

Thank you for your time and consideration!

seebeen:
Let's use this as a starting point. Looking forward to your comments @olivervogel 

Will close #1239


olivervogel:
Wow, that was fast. I will take a look later. 👀

tests/images/softuni.png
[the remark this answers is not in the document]
olivervogel: I guess we can use this one: https://github.com/Intervention/image/blob/develop/tests/images/trim.png

seebeen:
> Works very well. I have only made a few minor comments about the PSR standard. Otherwise you can try to integrate the modifier with as good as it gets same functionality for the GD library, then I can merge it.

I avoid using PHPStan as the plague. Do you mind if I add-in PHPCS as a dev dependency with a PSR12 standard config? We can spend some time tweaking it afterward :)

tests/images/softuni.png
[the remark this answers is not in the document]
seebeen: No problems.

olivervogel:
> I avoid using PHPStan as the plague. Do you mind if I add-in PHPCS as a dev dependency with a PSR12 standard config?

Do you want to add phpcs additionally or instead of phpstan?

seebeen:
> > I avoid using PHPStan as the plague. Do you mind if I add-in PHPCS as a dev dependency with a PSR12 standard config?
> 
> Do you want to add phpcs additionally or instead of phpstan?

Add it as an additional dev dependency - wouldn't want to impose my workflow setup on you (or anyone else). People who use both - will get both. Others (like me) get to pick a camp 😊
Personally I use VSCode and Intelephpense - with custom phpcs rules.

olivervogel:
> Add it as an additional dev dependency

Ok sure. Create a new PR for this if you want.

seebeen:
New PR created - #1275.

src/Image.php
```diff
@@ -749,6 +750,15 @@ public function place(
         return $this->modify(new PlaceModifier($element, $position, $offset_x, $offset_y));
     }
 
+    /**
+     * {@inheritdoc}
+     *
+     * @see ImageInterface::trim()
+     */
+    public function trim(int $tolerance = 0): ImageInterface {
```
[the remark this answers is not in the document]
seebeen: Fixed

seebeen:
Added the GD trim modifier. Converting this to proper PR, and awaiting your feedback 💪

src/Drivers/Gd/Modifiers/TrimModifier.php
seebeen, line 1: We need to use corner color detection because threshold doesn't work with IMG_CROP_SIDES.

src/Drivers/Gd/Modifiers/TrimModifier.php
```diff
@@ -0,0 +1,59 @@
...
+            [$width - 1,$height - 1],
+        ];
+
+        foreach ($corners as $corner) {
+            $thisColor = imagecolorat($gd, $corner[0], $corner[1]);
+
+            $rgb    = imagecolorsforindex($gd, $thisColor);
+            $red   += round(round(($rgb['red'] / 0x33)) * 0x33);
+            $green += round(round(($rgb['green'] / 0x33)) * 0x33);
+            $blue  += round(round(($rgb['blue'] / 0x33)) * 0x33);
```
olivervogel, line 42: Just curious. Why do you use `0x33` instead of dividing directly by 51?

src/Modifiers/TrimModifier.php
```diff
@@ -0,0 +1,28 @@
...
+     * Class constructor
+     *
+     * @param int $tolerance Tolerance level for trim operation
+     */
+    public function __construct(
+        /**
+         * Tolerance level for trim operation
+         *
+         * @var int
+         */
```
olivervogel, line 24: This is already described above.

src/Drivers/Gd/Modifiers/TrimModifier.php
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        $red   /= 4;
+        $green /= 4;
+        $blue  /= 4;
+
+        $trimColor = imagecolorallocate($gd, (int)$red, (int)$green, (int)$blue);
+        $trimmed   = imagecropauto($gd, IMG_CROP_THRESHOLD, $this->tolerance, $trimColor);
```
olivervogel, line 50: Basically it works. However, above certain tolerance values the function returns `false`. I think this should be handled.  For example, the test image "trim.png" shows an error from tolerance `67`.

I think it should be possible, as with Imagick, to enter values from 0 to 100 without errors occurring.

If a very high tolerance value is entered in Imagick, the result is a 1 x 1 pixel image.

src/Drivers/Gd/Modifiers/TrimModifier.php
seebeen: Because RGB values are in hexadecimal notation. I thought it made sense 😅.

src/Drivers/Gd/Modifiers/TrimModifier.php
seebeen: Do you want to limit the tolerance to 50?

src/Modifiers/TrimModifier.php
seebeen: Agreed

src/Drivers/Gd/Modifiers/TrimModifier.php
olivervogel: > Do you want to limit the tolerance to 50?

No, I would like to be able to set the parameter value between 0 and 100 and get the same result in GD and Imagick. Of course this is not 100% possible, but we should get as close as possible.

olivervogel: > Meanwhile - do we really need to place limitations, and safeguards? I'm more inclined to set no limits for tolerance, and add a note in userdocs. How people choose to use the lib - is up to them?

Yes. I agree.

seebeen: Do you want to merge as is, I can fill in some more extreme use cases in the following days? :)

seebeen: Ok, let me know what do you want me to do :)

olivervogel: > Ok, let me know what do you want me to do :)

Intervention Image is a standardized API for GD and Imagick. The main goal is to make the results of the API independent of the driver used and thus ensure interchangeability.

Of course, this goal is not always 100% achievable, and especially in this case it will not work, because both drivers work completely different. However, I have always done my best to get as close to 100% as possible. I expect the same for merge requests.

We have a new method that takes a parameter. The parameter can be passed values from 0 to 100. 

- The result of each of these values should be an instance of ImageInterface.
- The modification of the image should be as similar as possible with both drivers, same images and parameters.

I think the output of `trim()` with Imagick can be used as a model.

If you have any questions, feel free to contact me.

olivervogel:
I wanted to ask if you still plan to continue here? If you have any questions or uncertainties, please get in touch. If you're no longer interested, that's fine too, of course. I won't ask any more questions and will close the PR.

seebeen:
Heya. 

Yes, I do want to continue. I'll aim to finalize by monday.

olivervogel:
I took the basis of @seebeen  as a starting point and further developed the feature in [this PR](https://github.com/Intervention/image/pull/1322). It continues there and I am closing this one.

PR #1275, opened 2024-01-17
Added PHPCS config and dependencies

seebeen:
Implemented PSR-12 PHPCS config and added .editorconfig for development standardization

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+# editorconfig.org
+
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.php]
+indent_size = 4
--- a/composer.json
+++ b/composer.json
@@ -1,45 +1,45 @@
 {
-    "name": "intervention/image",
-    "description": "PHP image manipulation",
-    "homepage": "https://image.intervention.io/",
-    "keywords": ["image", "gd", "imagick", "watermark", "thumbnail", "resize"],
-    "license": "MIT",
-    "authors": [
-        {
-            "name": "Oliver Vogel",
-            "email": "oliver@intervention.io",
-            "homepage": "https://intervention.io/"
-        }
-    ],
-    "require": {
-        "php": "^8.1",
-        "ext-mbstring": "*",
-        "intervention/gif": "^4"
-    },
-    "require-dev": {
-        "phpunit/phpunit": "^9",
-        "mockery/mockery": "^1.6",
-        "phpstan/phpstan": "^1",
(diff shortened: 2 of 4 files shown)


composer.json
```diff
@@ -19,7 +19,9 @@
     "require-dev": {
         "phpunit/phpunit": "^9",
         "mockery/mockery": "^1.6",
-        "phpstan/phpstan": "^1"
+        "phpstan/phpstan": "^1",
+        "dealerdirect/phpcodesniffer-composer-installer": "^1.0",
```
olivervogel, line 23: Why do we need this package?

phpcs.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0"?>
+<ruleset name="Intervention/image">
+
+  <rule ref="PSR12"/>
+
+  <rule ref="Generic.Files.LineLength">
+    <severity>0</severity>
+  </rule>
+
+  <file>src/</file>
```
olivervogel, line 10: Folder `tests` should also be included.

.editorconfig
```diff
@@ -0,0 +1,14 @@
+# editorconfig.org
```
olivervogel, line 1: `.editorconfig` should be excluded via `.gitattributes`

composer.json
seebeen: To automatically setup phpcs for various IDEs - so our phpcs.xml config works OOB

phpcs.xml
seebeen: Roger roger

.editorconfig
seebeen: Done

seebeen:
Changes implemented per request.

composer.json
olivervogel: I tried to figure out what this package does but still don't get it.

> This composer installer plugin allows for easy installation of PHP_CodeSniffer
> No more symbolic linking of directories, checking out repositories on specific locations or changing the phpcs configuration.

Why do I need an installation package for "PHP_CodeSniffer" when you can install the tool without any other dependencies? `phpcs.xml` is read in automatically and/or can be integrated via the GitHub testing workflow.

Don't get me wrong. I don't want to criticize here, but always try to keep package dependencies as low as possible.

I think the IDE setup belongs on the machine of the developer. In my opinion, we can do without this dependency in this case.

seebeen: OK, I'll remove it.

seebeen:
And again 😅

olivervogel:
Thank you very much. I included the code standard validation additionally in the Github Workflow and added validation of line length maximum. Had to tweak 3 lines for the checks to pass.

seebeen:
Awesome. Continuing with the Trim implementation.

PR #1286, opened 2024-01-29
Fix ImageInterface encode() method argument

edgarsn:
ImageInterface encode argument is not aligned with Image class.

**ImageInterface.php**

`public function encode(EncoderInterface $encoder): EncodedImageInterface;`

**Image.php**

`public function encode(EncoderInterface $encoder = new AutoEncoder()): EncodedImageInterface`

This is causing static analysis tools to fail as well as confuse IDEs with error:

`Method Intervention\Image\Interfaces\ImageInterface::encode() invoked with 0 parameters, 1 required.`

### DIFF
--- a/src/Interfaces/ImageInterface.php
+++ b/src/Interfaces/ImageInterface.php
@@ -66,7 +66,7 @@ interface ImageInterface extends IteratorAggregate, Countable
      * @param EncoderInterface $encoder
      * @return EncodedImageInterface
      */
-    public function encode(EncoderInterface $encoder): EncodedImageInterface;
+    public function encode(EncoderInterface $encoder = new AutoEncoder()): EncodedImageInterface;
 
     /**
      * Save the image to the specified path in the file system. If no path is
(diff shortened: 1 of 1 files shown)


src/Interfaces/ImageInterface.php
```diff
@@ -66,7 +66,7 @@ public function size(): SizeInterface;
      * @param EncoderInterface $encoder
      * @return EncodedImageInterface
      */
-    public function encode(EncoderInterface $encoder): EncodedImageInterface;
+    public function encode(EncoderInterface $encoder = new AutoEncoder()): EncodedImageInterface;
```
olivervogel, line 69: This refers to `Intervention\Image\Interfaces\AutoEncoder`, which does not exist.

edgarsn: Sorry about that, monday.. Fixed in the last commit.

olivervogel:
Thank you.

PR #1291, opened 2024-02-02
Remove redundant DrawableInterface for Rectangle class

miguilimzero:
`DrawableInterface` is already implemented by Polygon and it is not needed on Rectangle. *Just found this little detail while checking the methods for the documentation update.*


olivervogel:
Thank you. How did you find something like that?

PR #1214, opened 2023-06-26
Add throws annotation related to Image manipulation

VincentLanglet:
Hi @olivervogel,

I recently used this library and got exceptions when calling some methods but most of them were undocumented.
Keeping a `@throws` phpdoc up to date is useful
- For the user to know which method can thrown exception, and which exception are thrown
- For PHPStorm to report uncaught exceptions
- For static analysis tools (like PHPStan) to also report uncaught exceptions

I tried to add some on Image manipulation (the methods I'm using so far).
Is it ok to merge such phpdoc update ?
Thanks a lot.

PR #1303, opened 2024-02-28
Add Decoder and Encoder thrown exception

VincentLanglet:
Hi @olivervogel, as suggested in https://github.com/Intervention/image/pull/1214#issuecomment-1848577030, I'm now using the version 3 of intervention/image. But there is still no phpdoc about possibly thrown exception.

Keeping a `@throws` phpdoc up to date is useful
- For the user to know which method can thrown exception, and which exception are thrown
- For PHPStorm to report uncaught exceptions
- For static analysis tools (like PHPStan) to also report uncaught exception

Since there is way too much exception in this tool, I started with the most frequent: EncoderException and DecoderException.

In my case this would be helpful to tell PHPStan/Phpstorm that I need to try/catch
- ImageInterface::encode method
- ImageInterface::* modifiers methods
- ImageManager::read method

If you're interested, I can continue
- with other exception
- with a PR introducing the check done by phpstan (https://phpstan.org/blog/bring-your-exceptions-under-control)

I just recommend this to be in another PR to minimize the size.

### DIFF
--- a/phpstan.dist.neon
+++ b/phpstan.dist.neon
@@ -1,0 +1,20 @@
+parameters:
+    level: 4
+    paths:
+        - src
+    exceptions:
+        check:
+            missingCheckedExceptionInThrows: true
+        uncheckedExceptionClasses:
+            - Intervention\Image\Exceptions\AnimationException
+            - Intervention\Image\Exceptions\ColorException
+            - Intervention\Image\Exceptions\DriverException
+            - Intervention\Image\Exceptions\GeometryException
+            - Intervention\Image\Exceptions\FontException
+            - Intervention\Image\Exceptions\InputException
+            - Intervention\Image\Exceptions\NotSupportedException
+            - Intervention\Image\Exceptions\NotWritableException
+            - ImagickException
+            - ImagickDrawException
+            - ImagickPixelException
+            - Error
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace Intervention\Image;
 
-use Intervention\Image\Exceptions\RuntimeException;
 use Intervention\Image\Interfaces\CollectionInterface;
 use ArrayIterator;
 use Countable;
@@ -198,10 +197,6 @@ class Collection implements CollectionInterface, IteratorAggregate, Countable
      */
     public function slice(int $offset, ?int $length = null): CollectionInterface
     {
-        if ($offset >= count($this->items)) {
-            throw new RuntimeException('Offset exceeds the maximum value.');
-        }
-
         $this->items = array_slice($this->items, $offset, $length);
 
         return $this;
--- a/src/Drivers/Gd/Driver.php
+++ b/src/Drivers/Gd/Driver.php
@@ -6,7 +6,7 @@ namespace Intervention\Image\Drivers\Gd;
 
 use Intervention\Image\Drivers\AbstractDriver;
 use Intervention\Image\Exceptions\DecoderException;
-use Intervention\Image\Exceptions\RuntimeException;
+use Intervention\Image\Exceptions\DriverException;
 use Intervention\Image\Image;
 use Intervention\Image\Interfaces\ColorInterface;
 use Intervention\Image\Interfaces\ColorProcessorInterface;
@@ -36,7 +36,7 @@ class Driver extends AbstractDriver
     public function checkHealth(): void
     {
         if (!extension_loaded('gd') || !function_exists('gd_info')) {
-            throw new RuntimeException(
+            throw new DriverException(
                 'GD PHP extension must be installed to use this driver.'
             );
         }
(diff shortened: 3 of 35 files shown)


olivervogel:
Hey @VincentLanglet , thanks for your PR. The PHPDoc tags for the different exception types could be really helpful.

> If you're interested, I can continue

Absolutely, any help is very welcome. 

I only have one request. Even if some see it as nitpicking, I would suggest keeping the `@return` tag as the last item in the block. Apart from that, it already looks wonderful.

VincentLanglet:
> I only have one request. Even if some see it as nitpicking, I would suggest keeping the `@return` tag as the last item in the block. Apart from that, it already looks wonderful.

Sure, it's done in https://github.com/Intervention/image/pull/1303/commits/ba39c13d2b264c815026f39829dae00b0d818c63 commit.

> Hey @VincentLanglet , thanks for your PR. The PHPDoc tags for the different exception types could be really helpful.

This can be checked by PHPStan, I added the config in https://github.com/Intervention/image/pull/1303/commits/1ed31f3cdf48043ee3ca799d5d6af9c23cdec203#diff-a4ce6ffd8cb697aecebba7af3c784ed29078f645544c069ec7fecbe9417e2a04

> > If you're interested, I can continue
> 
> Absolutely, any help is very welcome.

I propose to merge this PR avoiding having a too big change (which is already not small) and I'll do same for others exceptions in another pr if it's ok for you @olivervogel

src/Collection.php
```diff
@@ -198,10 +197,6 @@ public function empty(): CollectionInterface
      */
     public function slice(int $offset, ?int $length = null): CollectionInterface
     {
-        if ($offset >= count($this->items)) {
-            throw new RuntimeException('Offset exceeds the maximum value.');
```
VincentLanglet, line 202: `array_slice([], 2, 1)` works and return `[]`.

I propose to allow this in order to avoid throwing exception in slice method (or you'll need to add `@throws RuntimeException` on every other method using slice).

src/Drivers/Gd/Driver.php
```diff
@@ -36,7 +36,7 @@ public function id(): string
     public function checkHealth(): void
     {
         if (!extension_loaded('gd') || !function_exists('gd_info')) {
-            throw new RuntimeException(
+            throw new DriverException(
```
VincentLanglet, line 39: I think it's better having a specific exception for such error, this would allows for example to ignore the check of this exception for phpstan or allow the user to have a specific behavior for them.

It BC because DriverException extends RuntimeException.

src/Drivers/Gd/Encoders/GifEncoder.php
```diff
@@ -43,7 +43,11 @@ protected function encodeAnimated(ImageInterface $image): EncodedImage
             );
         }
 
-        $builder->setLoops($image->loops());
+        try {
+            $builder->setLoops($image->loops());
```
VincentLanglet, line 47: setLoops may throw an exception if the loop is empty.

I try catch to transform this to a custom exception (and respect the `@throws` phpdoc)

src/Drivers/Imagick/Core.php
```diff
@@ -98,10 +97,6 @@ public function empty(): CollectionInterface
      */
     public function slice(int $offset, ?int $length = null): CollectionInterface
     {
-        if ($offset >= $this->count()) {
```
VincentLanglet, line 101: `array_slice([], 2, 1)` works and return `[]`.

I propose to allow this in order to avoid throwing exception in slice method (or you'll need to add `@throws RuntimeException` on every other method using slice).

phpstan.dist.neon
```diff
@@ -0,0 +1,20 @@
+parameters:
+    level: 4
+    paths:
+        - src
+    exceptions:
+        check:
+            missingCheckedExceptionInThrows: true
+        uncheckedExceptionClasses:
```
VincentLanglet, line 8: These exception arenot checked yet but could be in following pr

src/Exceptions/AnimationException.php
```diff
@@ -4,6 +4,6 @@
 
 namespace Intervention\Image\Exceptions;
 
-class AnimationException extends \RuntimeException
+class AnimationException extends RuntimeException
```
VincentLanglet, line 7: This is better because it allows to catch all Exceptions with
```
catch (Intervention\Image\RuntimeException)
```
rather than
```
catch (\RuntimeException)
```
which is too generic.

src/Collection.php
olivervogel: This is fine for the rather generic class `Collection`. However, it has an effect on the function `Image::sliceAnimation()`, which throws rather incomprehensible exceptions by GD/Imagick, which are rather confusing for someone who does not know the library inside out.

For this reason, I have reintegrated the check, but moved it to the `SliceAnimationModifier` classes.

olivervogel:
This all looks good. I have made a few minor adjustments and otherwise adopted it as it is.

PR #1306, opened 2024-02-29
Add more `@throws` annotation

VincentLanglet:
Hi @olivervogel, as promise, this is a follow up of https://github.com/Intervention/image/pull/1303.

I just uncommented the exception from phpstan config and followed the errors. It seems that a lot of methods are possibly throwing exception, and the `@throws RuntimeException` spread from method to method.

I stop so far because I'd like to know if you will be interested by such PR. I think it's debatable.
- Unlike Encoder/Decoder exceptions, most of the others exceptions (like colorexception, inputexception, geometryexception, ...) are kinda user-input related exception and shouldn't occurs when the user is using the right parameters. So we might consider they are not meant to be caught and doesn't need to be fully documented.
- On the otherside you might prefer to have all the exception documented without questions.

Your choice.

### DIFF
--- a/phpstan.dist.neon
+++ b/phpstan.dist.neon
@@ -8,7 +8,6 @@ parameters:
         uncheckedExceptionClasses:
             - Intervention\Image\Exceptions\AnimationException
             - Intervention\Image\Exceptions\ColorException
-            - Intervention\Image\Exceptions\DriverException
             - Intervention\Image\Exceptions\GeometryException
             - Intervention\Image\Exceptions\FontException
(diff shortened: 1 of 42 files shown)


olivervogel:
Great, thank you very much for your efforts, @VincentLanglet . The changes can be merged when the failing checks in the pipeline are fixed.

VincentLanglet:
> Great, thank you very much for your efforts, @VincentLanglet . The changes can be merged when the failing checks in the pipeline are fixed.

Done

PR #1308, opened 2024-03-01
Add more exception annotation

VincentLanglet:
Last PR on the "exception" track @olivervogel.

NB: `@throws void` is a tips to tell phpstan than even if the method is technically throwing exception, it does throws nothing in this case.

### DIFF
--- a/phpstan.dist.neon
+++ b/phpstan.dist.neon
@@ -6,8 +6,6 @@ parameters:
         check:
             missingCheckedExceptionInThrows: true
         uncheckedExceptionClasses:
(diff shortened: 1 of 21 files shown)


olivervogel:
Thank you.

VincentLanglet:
> Thank you.

You're welcome ; does a new release would be possible ?

olivervogel:
> You're welcome ; does a new release would be possible ?

The project follows a monthly release cycle for minor releases. The next one will on March 16, 2024.

PR #1313, opened 2024-03-17
Avoid warning on getimage call

VincentLanglet:
Hi @olivervogel 

I'm getting some 
```
Warning: getimagesize(): PNG file corrupted by ASCII conversion
```
from this code.

Since, a `!in_array` check is thrown with an exception. The warning should be muted.

### DIFF
--- a/src/Drivers/Gd/Decoders/AbstractDecoder.php
+++ b/src/Drivers/Gd/Decoders/AbstractDecoder.php
@@ -18,7 +18,7 @@ abstract class AbstractDecoder extends GenericAbstractDecoder
      */
     protected function getMediaTypeByFilePath(string $filepath): string
     {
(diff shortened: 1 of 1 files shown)


olivervogel:
Thanks for your PR. Can you provide an image file that produces this warning? I would like to take a closer look at the reasons for this.

VincentLanglet:
> Can you provide an image file that produces this warning? I would like to take a closer look at the reasons for this.

Unfortunately `Warning: getimagesize(): PNG file corrupted by ASCII conversion` is a log from my production, and since the user got an exception he didn't imported the image... so I don't have any example :/

olivervogel:
>  so I don't have any example :/

That's ok. The changes still make sense.

PR #1318, opened 2024-03-22
Update TextBlock.php

soheylfarzane:
Fix Multi Line Text reverse in Arabic And Persian Text Bug If Text Is Arabic Or Persian Write Lines True


olivervogel:
Closing due to inactivity.

PR #1331, opened 2024-04-20
Up PHPStan level to 5

szepeviktor:
Includes a return type change.


olivervogel:
Thank you for your efforts. I like the idea of raising the overall level of phpstan. However, I think that no further ignore statements should be introduced. I think this is only doable with refactoring the code.

szepeviktor:
> I think this is only doable with refactoring the code.

I agree.

Issue #1324, opened 2024-03-26
Unable to mock final classes


clementlanot:
**Describe the bug**
Impossible to test with PHPUnit and Laravel.

**Code Example**
```
Image::shouldReceive('save')->andThrow(new NotWritableException());
```

Throw the error:
```
Mockery\Exception: The class \Intervention\Image\ImageManager is marked final and its methods cannot be replaced. Classes marked final can be passed in to \Mockery::mock() as instantiated objects to create a partial mock, but only if the mock is not subject to type hinting checks.
```

And using Mockery:
```
$m = Mockery::mock(new Image());
```
Throw the error:
```
Cannot redeclare Mockery_3_Intervention_Image_Laravel_Facades_Image::shouldReceive()
```

**Expected behavior**
Perhaps we should not declare classes as final?

**Environment (please complete the following information):**
 - PHP Version: 8,2
 - Laravel 11
 - Intervention Image Version: 3

olivervogel:
I think this is more a topic for Mockery. Also have a look at [dg/bypass-finals](https://github.com/dg/bypass-finals).

ChrGriffin:
_Strongly_ disagree that this is an issue for Mockery to resolve. Mockery is subject to the rules of how the language operates, like anyone else. On the one hand, you could simply _not_ mark the class as `final`. On the other hand, Mockery could try some hacky nonsense to circumvent the basic rules of the language. It doesn't take a genius to see which option is preferable.

There's no real reason to mark the ImageManager as `final` to begin with. Literally the only thing you accomplish by doing so is cause a pain for those who need to mock it for testing or extend it. Sure, there are ways around it, like the package you linked. But why offload the responsibility on us when there's no good reason for you not to make a minor change to the package?

Heck, I'd make the PR myself and save you the work (such as it is), if it wasn't clear you aren't disposed towards it.

PR #1336, opened 2024-04-26
Add ImageManager interface

adamaveray:
In v3 the ImageManager class is now final making it impossible to mock out in tests, and while most other classes in this library have corresponding interfaces enabling mocking this class does not.

This PR adds a new ImageManagerInterface interface with the 4 public instance methods from ImageManager. I've based it off how the Image & ImageInterface pair is set up, so the docblock contents for those methods have also been moved to the interface.

Addresses #1324.

### DIFF
--- a/src/ImageManager.php
+++ b/src/ImageManager.php
@@ -4,14 +4,14 @@ declare(strict_types=1);
 
 namespace Intervention\Image;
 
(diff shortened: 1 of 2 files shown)


olivervogel:
Thank you, I think this can be adopted as is.

Issue #1347, opened 2024-05-05
DrawLine with unwanted colors (Imagick)


olivervogel:
**Describe the bug**
If you use the DrawLineModifer with the Imagick driver and draw a (semi) transparent line, an unwanted color is formed along the line.

**Code Example**

```php
$image = $manager
    ->create(500, 500)
    ->fill('ff5500')
    ->drawLine(function (LineFactory $line) {
        $line->color('fff4');
        $line->from(100, 100);
        $line->to(400, 100);
        $line->width(40);
    });
```

**Expected behavior**
Draw line without unwanted colors.

**Images**
<img width="499" alt="Bildschirmfoto 2024-05-05 um 13 10 37" src="https://github.com/Intervention/image/assets/884642/b64d6a7c-c2ba-4820-b2c8-2a2997fca3c4">

**Environment:**
 - PHP Version: PHP 8.3
 - OS: MacOS
 - Intervention Image Version: 3.6.3
 - GD or Imagick: Imagick

PR #1348, opened 2024-05-05
Fix bug with unwanted color in DrawLineModifer

olivervogel:
See #1347

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -89,7 +89,10 @@ jobs:
       - name: Install dependencies
         run: composer update --${{ matrix.stability }} --prefer-dist --no-interaction
 
-      - name: Which Imagick Version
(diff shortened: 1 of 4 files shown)


olivervogel:
The tests are running in the GitHub workflow, but for unknown reasons there is an error in the Docker test environment.

```bash
docker-compose run --build --rm tests

[...]

There was 1 failure:
                                                                                  
1) Intervention\Image\Tests\Unit\Drivers\Gd\Modifiers\DrawLineModifierTest::testApplyTransparent
Failed asserting that color rgba(255, 255, 255, 255) equals rgba(255, 129, 66, 255)
Failed asserting that an array contains 255.

/project/tests/BaseTestCase.php:79
/project/tests/Unit/Drivers/Gd/Modifiers/DrawLineModifierTest.php:36
```

PR #1353, opened 2024-05-23
Silence php warning when throwing decoder exceptions

VincentLanglet:
Hi @olivervogel, 

I'm getting multiple warnings:
```
Warning: imagecreatefromjpeg(): gd-jpeg error: cannot allocate gdImage struct
```
or similar message when decoding bad inputs.

Since parent::decode call is already throwing an exception
```
throw new DecoderException('Unable to decode input')
```
there is no need to keep such warning which cannot be easily silenced.


olivervogel:
Yeah, that sounds good. I think this can be merged as is. Although I wonder why the [checks fail](https://github.com/Intervention/image/actions/runs/9209636419/job/25338134444#step:16:13) since you haven't changed anything in the files mentioned.

olivervogel:
> Do you want me to fix them ?

Would be great if you could take a look.

VincentLanglet:
Fixed @olivervogel

olivervogel:
Thanks alot.

PR #1354, opened 2024-05-28
Update doctype ImageManagerInterface

Tsuguya:
The [example](https://image.intervention.io/v3/basics/instantiation#examples) uses an array of strings.
It seems to be fine either way when you look at it.

https://github.com/Intervention/image/blob/267d588eb100d5cf36c91fffa3642087aa4c1028/src/Drivers/AbstractDriver.php#L69-L79


olivervogel:
Yes, you're right. Thank you for your attention.

PR #1372, opened 2024-07-02
Handle non-standard but still in use image/x-jpeg mimetype as JPEG

dvlpp:
I stumbled upon this issue for the second time recently, so I decided to try to submit a PR: I sometimes have to deal with image/x-jpeg files (seems that it's a non-standard mime type, but some use it...).

Without this PR, the code breaks with
`"image/x-jpeg" is not a valid backing value for enum Intervention\Image\MediaType`


olivervogel:
Thanks for the PR. You're right, the non-standard type should be included. Can you please fix the erroneous tests, then I can merge this as is.

olivervogel:
Thanks.

PR #1375, opened 2024-07-03
Allow animated WEBP's

phknot:
The current logic merges all the layers when using WEBP and Imagick. In an animated image, these layers are the separate frames. Not merging the layers when we're dealing with an animation allows for the creation of animated webp files.

Note: this still does only work with Imagick, GD has no animated webp support yet.

### DIFF
--- a/src/Drivers/Imagick/Encoders/WebpEncoder.php
+++ b/src/Drivers/Imagick/Encoders/WebpEncoder.php
@@ -21,7 +21,10 @@ class WebpEncoder extends GenericWebpEncoder implements SpecializedInterface
         $imagick = $image->core()->native();
         $imagick->setImageBackgroundColor(new ImagickPixel('transparent'));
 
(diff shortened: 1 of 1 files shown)


src/Drivers/Imagick/Encoders/WebpEncoder.php
```diff
@@ -21,7 +21,10 @@ public function encode(ImageInterface $image): EncodedImage
         $imagick = $image->core()->native();
         $imagick->setImageBackgroundColor(new ImagickPixel('transparent'));
 
-        $imagick = $imagick->mergeImageLayers(Imagick::LAYERMETHOD_MERGE);
+        if(! $image->isAnimated()) {
```
olivervogel, line 24:
```suggestion
        if (!$image->isAnimated()) {
```

olivervogel:
Thanks.

PR #1399, opened 2024-10-27
Fix typo in mime type.

ADmad:


ADmad:
I am a bit surprised this didn't cause any test failure. There does seem to be a test for it https://github.com/Intervention/image/blob/3618c8fbaf0e1019cb96dc41cc8ba811c5a8682e/tests/Unit/Drivers/Gd/Encoders/AvifEncoderTest.php#L22

olivervogel:
Thank you. There is a test, but during the test run the type is taken directly from the encoded image data via `mime_content_type()` and is correct. The string parameter that was passed incorrectly  is used to read the mime type from the instance of `EncodedImage::class` via without having to analyze the image data (with `mime_content_type()`), as the analysis functions require the fileinfo extension and I did not want to get into that dependency for this one method.

However, I will add tests for this case.

Thanks again.

olivervogel:
Is fixed in `3.9.1`.

Issue #1383, opened 2024-08-01
Update to PHPUnit 11


olivervogel:
## Describe the feature you'd like
Update to latest stable PHPUnit version 11. Requirement is PHP 8.2 minimum - currently Intervention Image only requires 8.1.

olivervogel:
[PHPUnit 11 is now integrated as an alternative to version 10.](2763c23bcd4b79d5872ea8b87ddaf34bdf7f6023)

PR #1401, opened 2024-10-27
Allow PHPUnit 11

ADmad:
Refs #1383

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "intervention/gif": "^4.2"
     },
     "require-dev": {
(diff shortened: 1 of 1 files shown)


olivervogel:
Are PHPUnit 10 and 11 the same to the extent that they can be exchanged without adjustments?

ADmad:
Yes they are quite similar, the main changes I have seen are deprecation of use of docblock annotations (you already use attributes) and changes to the mocking API (you use Mockery so unaffected by it).

PR #1415, opened 2025-01-10
Refactor imagick driver CropModifier::class

olivervogel:


src/Drivers/Imagick/Modifiers/CropModifier.php
```diff
@@ -14,80 +13,49 @@ class CropModifier extends GenericCropModifier implements SpecializedInterface
...
-                    0,
-                    ($this->offset_x * -1) + $originalSize->width() - $crop->pivot()->x() - 1,
-                    ($this->offset_y * -1) - $crop->pivot()->y() - 1,
-                );
-            }
+        return $image;
+    }
 
-            $frame->native()->drawImage($draw);
+    private function imagemagickMajorVersion(): int
```
deluxetom, line 53: I think you should move this to be a static function in the Imagick Driver

PR #1420, opened 2025-01-18
Strip parameter to encoders that support exif

deluxetom:
work in progress

### DIFF
--- a/src/Drivers/Imagick/Driver.php
+++ b/src/Drivers/Imagick/Driver.php
@@ -171,4 +171,21 @@ class Driver extends AbstractDriver
 
         return $matches['version'];
     }
+
+    /**
+     * @param Imagick $imagick
+     *
+     * @throws \ImagickException
+     * @return void
+     */
+    public static function stripExifKeepICCProfiles(Imagick $imagick): void
+    {
+        $profiles = $imagick->getImageProfiles('icc');
+
+        $imagick->stripImage();
+
+        if ($profiles !== []) {
+            $imagick->profileImage("icc", $profiles['icc']);
+        }
+    }
 }
--- a/tests/Unit/Drivers/Gd/Encoders/JpegEncoderTest.php
+++ b/tests/Unit/Drivers/Gd/Encoders/JpegEncoderTest.php
@@ -48,7 +48,6 @@ final class JpegEncoderTest extends GdTestCase
         $result = $encoder->encode($image);
         $this->assertMediaType('image/jpeg', $result);
         $this->assertEquals('image/jpeg', $result->mimetype());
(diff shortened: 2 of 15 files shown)


src/Drivers/Imagick/Driver.php
```diff
@@ -171,4 +171,21 @@ public static function version(): string
...
         return $matches['version'];
     }
+
+    /**
+     * @param Imagick $imagick
+     *
+     * @throws \ImagickException
+     * @return void
+     */
+    public static function stripExifKeepICCProfiles(Imagick $imagick): void
```
olivervogel, line 181: I think this fits better in `Intervention\Image\Drivers\Imagick\Core::class` and doesn't have to be static.

```php
$core = $core-> stripExifKeepICCProfiles();
```

tests/Unit/Drivers/Gd/Encoders/JpegEncoderTest.php
```diff
@@ -37,4 +37,17 @@ public function testEncodeProgressive(): void
         $this->assertEquals('image/jpeg', $result->mimetype());
         $this->assertTrue($this->isProgressiveJpeg($result));
     }
+
+    public function testEncodeStripExif(): void
```
olivervogel, line 41: The GD thing is tricky. There is no real function to delete meta data, but it is possible to create a new GDImage object and copy the data into it. The metadata is not transferred.

However, this does not lead to a consistent result. As the copy action described is used at various points in Modifier in order to achieve different results.

- src/Drivers/Gd/Modifiers/RotateModifier.php:83
- src/Drivers/Gd/Modifiers/ContainModifier.php:53
- src/Drivers/Gd/Modifiers/QuantizeColorsModifier.php:45
- src/Drivers/Gd/Modifiers/BlendTransparencyModifier.php:25
- src/Drivers/Gd/Modifiers/CropModifier.php:46
- src/Drivers/Gd/Modifiers/ResizeModifier.php:40
- src/Drivers/Gd/Modifiers/CoverModifier.php:40

Basically, the metadata is (almost) always lost with GD, even if this is not desired. :( This is also the reason why I have not yet integrated a “strip” feature.

olivervogel: Metadata is apparently not taken into account at all with GD.

```php
$data = fopen('php://temp', 'r+');
$img = imagecreatefromjpeg('exif.jpg'); // decode
imagejpeg($img, $data); // encode
$exif = exif_read_data($data); // exif gone
```

olivervogel:
I have already developed this further here: https://github.com/Intervention/image/pull/1421

PR #1468, opened 2026-02-08
Improve decoder exception messages with specific error descriptions

### LINKED ISSUE #1234: Decoder chain should output specific error messages instead of the generic "Unable to decode input"

QDenka:
## Summary

This PR replaces all generic `Unable to decode input` error messages across all decoders with specific, descriptive messages that indicate:

- **What type of input was expected** (e.g., `Expected a GdImage instance`, `Expected an SplFileInfo object`)
- **What went wrong** (e.g., `The binary data could not be read as an image`, `The image format of the file is not supported`)

## Problem

As described in #1234, when the decoder chain fails, the last decoder's generic `Unable to decode input` message is propagated — giving no indication of what went wrong or what input type was expected. This makes debugging difficult.

## Solution

Each decoder now throws a specific error message. Examples:

| Decoder | Old Message | New Message |
|---------|------------|-------------|
| FilePathImageDecoder | `Unable to decode input` | `Unable to decode input. The given path does not point to a readable file.` |
| BinaryImageDecoder | `Unable to decode input` | `Unable to decode input. The binary data could not be read as an image.` |
| HexColorDecoder | `Unable to decode input` | `Unable to decode input. The given string is not a valid hex color code.` |
| NativeObjectDecoder (GD) | `Unable to decode input` | `Unable to decode input. Expected a GdImage instance.` |
| NativeObjectDecoder (Imagick) | `Unable to decode input` | `Unable to decode input. Expected an Imagick instance.` |

All 25 decoder files across GD, Imagick, Color, and generic decoders have been updated.

Fixes #1234

### DIFF
--- a/src/Colors/Cmyk/Decoders/StringColorDecoder.php
+++ b/src/Colors/Cmyk/Decoders/StringColorDecoder.php
@@ -19,12 +19,12 @@ class StringColorDecoder extends AbstractDecoder implements DecoderInterface
     public function decode(mixed $input): ImageInterface|ColorInterface
     {
         if (!is_string($input)) {
-            throw new DecoderException('Unable to decode input');
+            throw new DecoderException('Unable to decode input. Expected a string.');
         }
 
         $pattern = '/^cmyk\((?P<c>[0-9\.]+%?), ?(?P<m>[0-9\.]+%?), ?(?P<y>[0-9\.]+%?), ?(?P<k>[0-9\.]+%?)\)$/i';
         if (preg_match($pattern, $input, $matches) != 1) {
-            throw new DecoderException('Unable to decode input');
+            throw new DecoderException('Unable to decode input. The given string is not a valid CMYK color string.');
         }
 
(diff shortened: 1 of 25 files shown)


olivervogel:
Thank you for your efforts. Unfortunately, the adjustments don't really help. There is a fundamental flaw in the decoder chain, as it “forgets” all previous exception messages when work continues in the chain.

Only the [last message is output at the end](https://github.com/Intervention/image/blob/develop/src/InputHandler.php#L104); all previous messages are no longer relevant, which poses a problem and your updated texts would never be displayed.

[I have already solved this issue by including an upfront check](https://github.com/Intervention/image/blob/next/src/InputHandler.php#L116-L121) to see whether this decoder is suitable for the input. Furthermore, there is an overall improvement of the detail in the exception messages and a logical tree of exception types.

Since this comes with breaking changes, it's currently only available in the [next branch](https://github.com/Intervention/image/tree/next), which will be the next major version 4 of this package, scheduled for the coming months.

PR #1484, opened 2026-05-02
Stop Imagick contrast from lifting every pixel

nlemoine:
`sigmoidalContrastImage` takes its midpoint in QuantumRange units, not normalized 0..1. Passing 0 anchors the curve at pure black, which lifts every pixel including the midtone. That's brightness, not contrast. GD's `IMG_FILTER_CONTRAST` is midtone-symmetric, so the same `$image->contrast(N)` gave different output across drivers. On `rgb(0, 174, 240)`: GD gives `rgb(0, 206, 255)`, Imagick gave `rgb(0, 252, 255)`.

Anchor the curve at `QUANTUM_RANGE / 2`. Imagick output now matches GD within 1 LSB.

Added testApplyPreservesMidtone: a pure-grey pixel must `survivecontrast(30)` unchanged. Bumped testApply expectation from `00fcff` to `00cffc`.

### DIFF
--- a/src/Drivers/Imagick/Modifiers/ContrastModifier.php
+++ b/src/Drivers/Imagick/Modifiers/ContrastModifier.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Intervention\Image\Drivers\Imagick\Modifiers;
 
+use Imagick;
 use ImagickException;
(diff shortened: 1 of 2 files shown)


olivervogel:
Thanks.

nlemoine:
Thank you @olivervogel !

PR #1485, opened 2026-05-10
Add Size::dimensions() for destructuring (fixes #1480)

### LINKED ISSUE #1480: Enable array destructuring of Size::class with width & height
## Describe the feature you'd like

Destructure a Size::class and be able to get width & height.

```php
[$width, $height] = $image->size(); // (int) $width, (int) $height
```

## Is your feature request related to a problem? Please describe.

Currently, a Size::class can be destructured, but you end up with individual points because the class is extended from Polygon::class. It would be better if only the width and height were destructured for this class like described above.

```php
// width & height are both Intervention\Image\Geometry\Point first two points of the four rectangle points
[$width, $height] = $image->size(); 
```

Abdooo2235:
## Summary

Closes #1480.

Adds `Size::dimensions(): array{0: int, 1: int}` so consumers can destructure
size into a `[width, height]` int pair:

```php
[$width, $height] = $image->size()->dimensions();
```

## Why not change `[$w, $h] = $image->size()` directly?

The literal syntax in the original issue would require changing
`ArrayAccess::offsetGet` on `Size`. That offset is currently the way
`Polygon` (Size's parent) exposes the four corner `Point` objects, and
`Size::setWidth()`/`setHeight()` rely on it internally
(`\$this[1]->setX(...)`, etc.). Repurposing those offsets to return ints
would be a breaking change for every caller and for the class's own setters.

`dimensions()` is the non-breaking alternative: same ergonomics for
destructuring, zero risk for existing code.

## Tests

Added to `tests/Unit/SizeTest.php`:

- `testDimensionsReturnsWidthAndHeightAsInts`
- `testDimensionsEnablesArrayDestructuring`
- `testDimensionsReflectsSetWidthAndSetHeight`
- `testArrayAccessStillReturnsCornerPoints` (explicit BC guard)

`vendor/bin/phpunit tests/Unit/SizeTest.php` passes locally (67 tests, 152 assertions).

## Notes

- `SizeInterface` updated to include the new method.
- No changes to `Polygon` or `ArrayAccess` semantics.

### DIFF
--- a/docs/superpowers/plans/2026-05-11-size-dimensions.md
+++ b/docs/superpowers/plans/2026-05-11-size-dimensions.md
@@ -1,0 +1,513 @@
+# Size::dimensions() Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Add `Size::dimensions(): array` so users can destructure size into `[int, int]` via `[$w, $h] = $image->size()->dimensions();`, resolving [Intervention/image#1480](https://github.com/Intervention/image/issues/1480) without breaking backwards compatibility.
+
+**Architecture:** Single new method on `Size` + matching interface entry on `SizeInterface`. No existing behaviour changes. Companion Claude CI workflow lives on a separate fork-only branch and is NOT included in the upstream PR.
+
+**Tech Stack:** PHP 8.3, PHPUnit 12 (attributes), PHPStan 2.1, PHP_CodeSniffer 4. Existing tests in `tests/Unit/SizeTest.php` already use `#[CoversClass(Size::class)]` and `BaseTestCase`.
+
+**Spec:** `docs/superpowers/specs/2026-05-11-size-dimensions-design.md`
+
+---
+
+## File Map
+
+| File | Action | Responsibility |
+|------|--------|----------------|
+| `src/Interfaces/SizeInterface.php` | Modify | Declare `dimensions(): array` |
+| `src/Size.php` | Modify | Implement `dimensions()` |
+| `tests/Unit/SizeTest.php` | Modify | Add 4 new test methods |
+| `.github/workflows/claude.yml` | Create (fork-only branch) | Claude GitHub Action — NOT in upstream PR |
+
+## Branch Strategy
+
+- Upstream PR branch: `fix/1480-size-dimensions` from `develop`. Contains library + test changes only.
+- Fork-only CI branch: `ops/claude-ci` from `main` (or `develop` if fork has no `main`). Contains only `.github/workflows/claude.yml`.
+
+---
+
+## Task 1: Set up the feature branch and verify baseline
+
+**Files:** none modified — branch setup only.
+
+- [ ] **Step 1: Confirm we are at the repo root and on `develop`**
+
+```bash
+cd C:/image
+git status
+git rev-parse --abbrev-ref HEAD
+```
+
+Expected: `develop` and a clean working tree (the `.claude/` untracked dir is fine; ignore it).
+
+- [ ] **Step 2: Create the feature branch off `develop`**
+
+```bash
+git checkout -b fix/1480-size-dimensions
+```
+
+Expected: `Switched to a new branch 'fix/1480-size-dimensions'`.
+
+- [ ] **Step 3: Install dependencies**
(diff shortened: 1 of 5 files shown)


olivervogel:
Thank you for the contribution and for taking the time to work on this.

Unfortunately, I can't merge the PR in its current form because the interface change introduce a breaking change.

I'd also prefer the method to be named `Size::toArray()`, as that would be more consistent with `Polygon::toArray()`. The current naming feels a bit inconsistent with the existing API conventions.

That said, there is also a larger issue with `toArray()` itself. Since the method is inherited from `Polygon::class` and currently used in [`src/Drivers/Gd/Modifiers/RotateModifier.php`](https://github.com/Intervention/image/blob/develop/src/Drivers/Gd/Modifiers/RotateModifier.php#L104) as a size to output individual points, so changing its behavior would have unintended side effects.

My actual goal is that one shouldn't have to call a method, but be able to destructure the object directly.

Thanks again.

Abdooo2235:
Thanks for the detailed feedback — that helps a lot.

Reading your three points, the architectural root seems to be that `Size extends Polygon`: that's what forces corner-Point semantics through `ArrayAccess` and blocks direct `[$w, $h] = $image->size()` destructuring, and it's also why `toArray()` can't be repurposed without breaking `RotateModifier`.

Would you accept a refactor where `Size` is decoupled from `Polygon` and implements `ArrayAccess<int, int>` directly — so `$size[0]` is width and `$size[1]` is height? Existing corner-Point access would move to an explicit method (e.g. `corners(): array<PointInterface>`), and the few internal callers (`RotateModifier::apply()`, `Size::setWidth/setHeight`, etc.) would be updated to use it. That gives you the destructuring API you described, keeps `Polygon::toArray()` untouched, and contains the BC impact to a clearly documented surface.

I'm happy to open a new PR with that scope if you'd consider merging it. Otherwise I'll close this out and wait for a future major release.

olivervogel:
I think decoupling Size from Polygon would be the best way to solve this. 

I have already tried that. However, I invested rather little time and noticed that it is more complex than can be done just like that.

I believe one issue actually is this area: https://github.com/Intervention/image/blob/develop/src/Drivers/Gd/Modifiers/RotateModifier.php#L88-L94

```php
$cutout = (new Size(
    imagesx($frame->native()),
    imagesy($frame->native()),
    $container->pivot()
))->alignHorizontally(Alignment::CENTER)
    ->alignVertically(Alignment::CENTER)
    ->rotate($this->rotationAngle());
```

A rectangular area is created from Size, but then methods are used that belong to Polygon. The blocker for simple decoupling is that size musts implement `rotate()` and the other methods even though they are not part of SizeInterface.

I think it makes sense to clean this up in the process. Maybe something like this would work:

```php
$cutout = Polygon::fromSize(new Size(
    imagesx($frame->native()),
    imagesy($frame->native()),
    $container->pivot()
))->alignHorizontally(Alignment::CENTER)
  ->alignVertically(Alignment::CENTER)
  ->rotate($this->rotationAngle());
```

This way the decoupled Size must not implement rotate() can be consumed by `imagefilledpolygon()` in https://github.com/Intervention/image/blob/develop/src/Drivers/Gd/Modifiers/RotateModifier.php#L104


I'm not sure atm whether that's the only part that's problematic.

I also noticed that Size is linked with Intervention\Image\Geometry\Rectangle. Unfortunately this whole arrangement is rather a design flaw and not ideal. I have a feeling this can not be cleaned up without breaking changes.

Abdooo2235:
Appreciate the detailed pointer — the `RotateModifier` chain was exactly the kind of hidden coupling I would have missed on a first pass.

`Polygon::fromSize()` as a factory makes sense; it keeps `Size` minimal and pushes the polygon behaviour to where it's actually needed. I'll also audit `Rectangle` and any other call sites that lean on the inheritance before settling on the final design.

I'll put together a draft PR with the decoupled `Size`, the factory, and the migrated call sites, and include a clear list of the BC breaks in the description so you can judge the impact. Going for a clean cut over a transition layer unless you'd prefer otherwise — easier to reason about, and you've already flagged that a major release seems appropriate.

I'll open the new PR (rather than reopen this one) so the conversation stays focused. No timeline promises, but I'll keep it scoped.

olivervogel:
I had some time and sketched something out: https://github.com/Intervention/image/pull/1487

Abdooo2235:
Beautiful — you built exactly what we were converging on, including the keyed-destructuring (`['width' => $w]`) variant which I hadn't proposed. I'll close this out and move my own conversation to #1487. Thanks for the speed.

PR #1487, opened 2026-05-13
Implement Size destructuring

### LINKED ISSUE #1480: Enable array destructuring of Size::class with width & height

olivervogel:
By decoupling Size from Polygon.

```php
[$width, $height] = $image->size();
['width' => $width, 'height' => $height] = $image->size();
```

Closes #1480


Abdooo2235:
Thanks for picking this up — the shape is exactly what I had in mind, and the keyed-destructuring is a nice addition.

One small thing on `Size::offsetGet`: the default branch throws `\ErrorException`. That's conventionally reserved for error-handler-converted PHP errors. For "unknown offset on a read-only ArrayAccess object," PHP's stdlib convention is `\OutOfBoundsException` (or `\InvalidArgumentException` — which is also what `Size`'s constructor uses for negative width/height in this codebase). Happy to send a small PR against your branch if useful.

Either way, looking forward to seeing this land.

olivervogel:
I see what you mean. I don't think `InvalidArgumentException` is quite the right choice here, since this isn't a function argument in the traditional sense and the exception is for semantically for bad function/method arguments.

I replaced it with `RuntimeException`, which matches the other `ArrayAccess` methods and keeps it in the libraries exception hierarchy.

PR #1499, opened 2026-06-19
Add 'image/bmp' case to BMP image processing switch

himtweety:
Add 'image/bmp' case to BMP image processing switch
    
    The `processBmp()` handler already supported various BMP MIME types (like
    `image/ms-bmp` and `image/x-bitmap`), but was missing the standard
    `image/bmp` identifier. Added the missing case condition.


olivervogel:
Intervention Image 2.x is EOL. Please use 4.x

PR #1501, opened 2026-06-21
Add 'image/bmp' case to BMP image processing switch

himtweety:
Add 'image/bmp' case to BMP image processing switch
    
    The `processBmp()` handler already supported various BMP MIME types (like
    `image/ms-bmp` and `image/x-bitmap`), but was missing the standard
    `image/bmp` identifier. Added the missing case condition.


olivervogel:
See https://github.com/Intervention/image/pull/1499

Please stop submitting the same same code as PR again and again.

himtweety:
sorry i did not see the message on other PR adn was under impression that the PR was not created due to some issue

PR #1506, opened 2026-07-08
Add JPEG XL (JXL) encoder

nlemoine:
This adds JPEG XL (JXL) as an encodable format.

It works like the HEIC encoder. Encoding goes through Imagick, and I wired JXL into the `Format`, `MediaType` and `FileExtension` enums. There's no JXL in GD, so the GD driver reports it as unsupported and throws `NotSupportedException`, same as HEIC today.

The encoder takes quality, strip and a lossless flag. For lossless I set the compression quality to 100, that's what ImageMagick's coder uses to switch libjxl to lossless mode (distance 0). See coders/jxl.c.

Putting image/jxl in the `MediaType` enum also lets the Imagick binary decoder read and round-trip .jxl input, since it gets the origin from `getImageMimeType()`. So reading works too, on any build that has libjxl.

I also added image/x-jxl next to image/jxl. It's not just an input alias. On one CI runner libmagic detected an encoded JXL blob as image/x-jxl while the others said image/jxl, so the test accepts both.

CI was the tricky part. libjxl-dev isn't packaged for ubuntu-22.04, it's only there from 24.04 on, so I moved the workflow to ubuntu-24.04. That brought out a second problem: 24.04 ships libheif 1.17, which moved its encoders into separate plugin packages, so libheif-dev on its own can't encode AVIF or HEIC anymore and those existing tests started failing with empty output. Installing the x265 and aomenc libheif plugins (plus the matching decoders) brings AVIF and HEIC back. I also bumped the ImageMagick cache key so the cached build gets rebuilt against libjxl.

One more thing on the tests. ImageMagick 6.9.13 doesn't build the JXL coder even when libjxl is there, only 7.1.2 does (you can see it in queryFormats). So the JXL encoder test and the Imagick JXL supports check skip when `Imagick::queryFormats('JXL')` is empty. They run on the 7.1.2 cells and skip on 6.9.13. I used queryFormats to stay consistent with how `Driver::supports()` already decides what's supported.

All six matrix cells are green. JXL runs for real on the three 7.1.2 cells (encode and lossless), and skips on the 6.9.13 cells.

Two things I'm not sure about, tell me what you prefer:

- image/x-jxl in the `MediaType` enum. I can drop it and keep only image/jxl, but then the media type check in the test has to tolerate whatever libmagic returns on a given runner.
- Moving CI to ubuntu-24.04. It was the simplest way to get libjxl, but it dragged in the libheif plugin change. If you'd rather stay on 22.04 I can try building libjxl another way, or just keep JXL out of CI.

Browser support is still niche for now. ~14-16%, Safari-only by default. But the [trend just turned](https://www.photoformatlab.com/blog/jpeg-xl-chrome-2026-image-format-guide): Chrome 145 (Feb 2026) re-added a Rust JXL decoder behind a flag, Firefox 152 (June 2026) put it in the release channel behind a Labs flag, and Chrome is expected to flip it on by default in H2 2026. Meaning it could jump to ~85-90% almost overnight. Having it now seems like the good time.

### DIFF
--- a/src/Encoders/JxlEncoder.php
+++ b/src/Encoders/JxlEncoder.php
@@ -13,13 +13,11 @@ class JxlEncoder extends SpecializableEncoder
      * Create new encoder object.
      *
      * @param null|bool $strip Strip EXIF metadata
-     * @param bool $lossless Encode losslessly, ignoring the quality value
      * @throws InvalidArgumentException
      */
     public function __construct(
         public int $quality = self::DEFAULT_QUALITY,
         public ?bool $strip = null,
-        public bool $lossless = false,
     ) {
         if ($quality < 0 || $quality > 100) {
             throw new InvalidArgumentException('Quality must be in range 0 to 100');
--- a/src/Encoders/JxlEncoder.php
+++ b/src/Encoders/JxlEncoder.php
@@ -1,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Intervention\Image\Encoders;
+
+use Intervention\Image\Drivers\SpecializableEncoder;
+use Intervention\Image\Exceptions\InvalidArgumentException;
+
+class JxlEncoder extends SpecializableEncoder
+{
+    /**
+     * Create new encoder object.
+     *
+     * @param null|bool $strip Strip EXIF metadata
+     * @param bool $lossless Encode losslessly, ignoring the quality value
+     * @throws InvalidArgumentException
+     */
+    public function __construct(
+        public int $quality = self::DEFAULT_QUALITY,
+        public ?bool $strip = null,
+        public bool $lossless = false,
+    ) {
+        if ($quality < 0 || $quality > 100) {
+            throw new InvalidArgumentException('Quality must be in range 0 to 100');
+        }
+    }
+}
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -7,7 +7,7 @@ permissions:
 
 jobs:
   run:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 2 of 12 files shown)


olivervogel:
Thanks. That's a really nice addition.

nlemoine:
@olivervogel I'll add the encoder to vips once it lands (vips needs the core Enums).

deluxetom:
@nlemoine thanks a lot! this is something I was looking forward to

src/Encoders/JxlEncoder.php
```diff
@@ -0,0 +1,28 @@
...
+     * Create new encoder object.
+     *
+     * @param null|bool $strip Strip EXIF metadata
+     * @param bool $lossless Encode losslessly, ignoring the quality value
+     * @throws InvalidArgumentException
+     */
+    public function __construct(
+        public int $quality = self::DEFAULT_QUALITY,
+        public ?bool $strip = null,
+        public bool $lossless = false,
```
olivervogel: One small thing: I'd like to remove the `lossless` option. 

Since JXL does not appear to have its own setting for lossless like [Webp](https://github.com/Intervention/image/blob/develop/src/Drivers/Imagick/Encoders/WebpEncoder.php#L66) does, this parameter is redundant and confusing because it suggests that `quality` and `lossless` can be controlled independently.

Everything else is great.

olivervogel (changes requested):

src/Encoders/JxlEncoder.php
nlemoine: Thanks for your review. You're right, I hesitated myself on that one. I just removed it.

deluxetom (approved):

nlemoine:
Thank you both @olivervogel & @deluxetom !

nlemoine:
@olivervogel The JXL vips encoder is ready, let me know when it's released.

olivervogel:
Just released [4.2.0](https://github.com/Intervention/image/releases/tag/4.2.0).

nlemoine:
Thanks! https://github.com/Intervention/image-driver-vips/pull/115

PR #1509, opened 2026-08-10
Color Palette Extraction

olivervogel:
The ability to generate color palettes from an image instance.

```php
// palette extraction
$popular = $image->colors()->popular(limit: 512);
$dominant = $image->colors()->dominant(limit: 5);
$theme = $image->colors()->theme(Intervention\Image\Colors\Theme::VIBRANT_MUTED);

// theme
$vibrant = $theme->vibrant;
$muted = $theme->muted;
$darkVibrant = $theme->darkVibrant;
$darkMuted = $theme->darkMuted;
$lightVibrant = $theme->lightVibrant;
$lightMuted = $theme->lightMuted;

// get color from palettes
$mostPopular = $popular->first();
$mostDominant = $dominant->first();

// color space transformation
$cmykPopular = $popular->toColorspace(Cmyk::class)
```

And more ...


deluxetom (commented):
Review focused on the findings with speed impact or bug potential (terse one-line style, one per location). Lower-severity cleanups omitted.

src/Analyzers/DominantPaletteAnalyzer.php
```diff
@@ -0,0 +1,322 @@
...
+    /**
+     * Analyze dominant colors in given image.
+     *
+     * @throws AnalyzerException
+     * @return array<RatedColor>
+     */
+    public function analyze(ImageInterface $image): array
+    {
+        // set seed for deterministic results
+        mt_srand(self::SEED);
```
deluxetom: L60: bug: `mt_srand(self::SEED)` reseeds PHP's process-global RNG and never restores it - after `dominant()`, every `mt_rand()`/`array_rand()`/`shuffle()` in the host app draws a fixed, predictable sequence (determinism and security hazard). Use a local `Random\Randomizer(new Random\Engine\Mt19937(self::SEED))` instead of the global engine.

src/Analyzers/DominantPaletteAnalyzer.php
```diff
@@ -0,0 +1,322 @@
...
+     * @throws AnalyzerException
+     * @return array<OklabColor>
+     */
+    private function initializeCentroids(array $pixels, int $k): array
+    {
+        $centroids = [];
+        $pixelIndices = array_keys($pixels);
+
+        // choose first centroid randomly (deterministic with seed)
+        $firstIndex = $pixelIndices[array_rand($pixelIndices)];
```
deluxetom: L167: bug: `array_rand($pixelIndices)` runs unconditionally - on a fully transparent image `collectColors()` yields nothing, so this throws an uncaught `ValueError` (while `popular()` on the same image returns an empty palette). Guard `$pixels === []` and return `[]`.

src/Colors/Palette.php
```diff
@@ -0,0 +1,244 @@
...
+    /**
+     * {@inheritdoc}
+     *
+     * @see PaletteInterface::first()
+     */
+    public function first(): ?ColorInterface
+    {
+        $colors = $this->colors;
+
+        return reset($colors);
```
deluxetom: L93: bug: `reset($colors)` returns `false` on an empty palette -> `TypeError` against `?ColorInterface` under strict_types (reachable via `popular()` on a fully transparent image). `last()` has the empty guard, `first()` doesn't; same bug in `Swatches::first()`. One line: `return $this->colors === [] ? null : $this->colors[array_key_first($this->colors)];`

src/Colors/Palette.php
```diff
@@ -0,0 +1,244 @@
...
+                fn(ColorInterface $color): ColorInterface => $color->toColorspace($sortColorspace),
+                $colors,
+            );
+        }
+
+        // sorting
+        usort(
+            $colors,
+            fn(ColorInterface $a, ColorInterface $b): int
+            => $b->channel($channel)->value() <=> $a->channel($channel)->value(),
```
deluxetom: L202: bug: the comparator forwards `$channel` unchanged into `ColorInterface::channel(string $classname)` - the advertised `ColorChannelInterface` object form always throws `TypeError` under strict_types (only the class-string form is tested). Also `$b <=> $a` sorts descending, so `sortByChannel()` is descending and `sortByChannelDesc()` ascending - both opposite their names and docblocks.

src/Colors/Palette.php
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @see PaletteInterface::slice()
+     */
+    public function slice(int $offset, ?int $length = null): PaletteInterface
+    {
+        $this->colors = array_slice($this->colors, $offset, $length);
```
deluxetom: L240: bug: `slice()` and `sortByChannel()` mutate `$this` and return it, while `toColorspace()` returns a clone - `$top = $palette->slice(0, 5)` silently discards the rest of the caller's palette. Pick one contract (cloning like `toColorspace()` avoids the data loss). Sorting also round-trips colors through integer-quantized colorspaces and writes the lossy values back, so the sorted palette's colors differ from the originals.

src/Colors/Swatches.php
```diff
@@ -0,0 +1,197 @@
...
+namespace Intervention\Image\Colors;
+
+use ArrayIterator;
+use Intervention\Image\Exceptions\ColorException;
+use Intervention\Image\Interfaces\ColorInterface;
+use Intervention\Image\Interfaces\ColorspaceInterface;
+use Intervention\Image\Interfaces\PaletteInterface;
+use Traversable;
+
+class Swatches extends Palette
```
deluxetom: L14: bug: Swatches inherits Palette's mutating `sortByChannel()`/`slice()`, whose colorspace round-trip unwraps `RatedColor` into plain colors and writes them back to `$this->colors` - the next `vibrant()` call crashes in `Classifier` reading `->rating`. Override them or don't extend Palette.

src/Colors/Swatches.php
```diff
@@ -0,0 +1,197 @@
...
+    public function lightMuted(): ?ColorInterface
+    {
+        return $this->classifier()->lightMuted();
+    }
+
+    /**
+     * @throws ColorException
+     * @return array<ColorInterface>
+     */
+    public function swatches(): array
```
deluxetom: L90: perf: no memoization - `count()`, `first()`, `last()`, `toArray()` and iteration each call `swatches()`, which rebuilds a fresh `Classifier` and a full 6-category pass every time; and `ColorExtractor::swatches()` uses `QuantizedPaletteAnalyzer(256^3)` -> identity quantization, so the palette holds every unique sampled color. A plain count + foreach + first() does ~18 full passes with repeated HSL conversions. Cache the six results once.

src/Drivers/Imagick/Analyzers/ColorCountAnalyzer.php
```diff
@@ -0,0 +1,29 @@
...
+     * {@inheritdoc}
+     *
+     * @see AnalyzerInterface::analyze()
+     *
+     * @throws AnalyzerException
+     */
+    public function analyze(ImageInterface $image): ?int
+    {
+        try {
+            return $image->core()->native()->getImageColors();
```
deluxetom: L24: perf: `getImageColors()` histograms every pixel of the full-resolution image just to decide the <=256-colors shortcut - on a 20-50MP photo this one call dominates `popular()`/`swatches()` runtime by orders of magnitude over the ~10k sampled pixels, defeating the sampling optimization on exactly the images it was designed for. Drop the pre-scan or run it on a downscaled copy.

src/Drivers/Gd/Analyzers/ColorCountAnalyzer.php
```diff
@@ -0,0 +1,23 @@
...
+class ColorCountAnalyzer extends GenericColorCountAnalyzer
+{
+    /**
+     * {@inheritdoc}
+     *
+     * @see AnalyzerInterface::analyze()
+     */
+    public function analyze(ImageInterface $image): ?int
+    {
+        $count = imagecolorstotal($image->core()->native());
```
deluxetom: L19: bug: `imagecolorstotal()` is always 0 for GD-decoded non-GIF images (the decoder converts palette images to truecolor), so the exact-color <=256 path only ever fires for GIFs on GD - the same PNG returns different color values on GD vs Imagick (e.g. pure red 255,0,0 becomes 249,6,6 on GD). The new tests assert only counts, which masks it.

deluxetom:
@olivervogel I also ran a review to find the biggest cuts:
```
- src/Colors/RatedColor.php:L43-181: yagni: implements ColorInterface via 15 pure delegation methods, yet every consumer unwraps ->coloranyway (ColorExtractor, Swatches). Plain{color, rating}value object withhash()/increaseRating(). -120.
- src/Colors/Swatches.php:L27-85: yagni: 6 one-line delegates to Classifier, a class with exactly one caller, rebuilt on every call. Merge Classifier into Swatches. -55.
- src/Colors/Palette.php:L160-187: shrink: 26-line match mapping every channel class to its colorspace. Channels live at X\Channels\Y, colorspace at X\Colorspace- derive via substr on::class + class_exists, 3 lines. -23.
- src/Colors/ColorExtractor.php:L32-56: shrink: popular() and dominant() duplicate the RatedColor→color array_map block. Move the mapping into extractColors(); both methods become return new Palette($this->extractColors(new ...($limit)));. -12.
- src/Colors/ColorExtractor.php:L68-72: delete: try/catch InvalidArgumentException around new QuantizedPaletteAnalyzer(256256256) - constructor only throws for limit < 1, a constant can't. -5.
- src/Colors/Quantizer.php:L127-131: delete: DivisionByZeroError catch in binIndexToNormalized - $levels >= 1enforced in constructor.return ($bin + 0.5) / $this->levels;, 1 line. -8.
- src/Colors/Quantizer.php:L112: delete: max(0.0, min(1.0, $value)) clamp immediately after throwing for the very same out-of-range values. Dead line.
- src/Analyzers/DominantPaletteAnalyzer.php:L189-193: delete: DivisionByZeroError catch around mt_rand() / mt_getrandmax() - mt_getrandmax() is never 0. -4.
- src/Colors/Classifier.php:L216-220: delete: DivisionByZeroError → RuntimeException → ColorException laundering chain. One if ($totalPopulation === 0) return null; guard at the top of findBestColor. -10 (also removes the catch at L646).
- src/Colors/Classifier.php:L152-160: stdlib: totalPopulation loop. array_sum(array_map(fn($c) => $c->rating, $this->colors)), 1 line. -5.
- src/Colors/Palette.php:L90-104 (+ Swatches L124-146): stdlib: copy-array-then-reset/end dance in first()/last(). array_key_first/array_key_lastone-liners (also fixes reset() returningfalseagainst?ColorInterface). -14.
- src/Analyzers/QuantizedPaletteAnalyzer.php:L16 + src/Analyzers/DominantPaletteAnalyzer.php:L18: delete: redundant implements AnalyzerInterface - parent AbstractPaletteAnalyzer already declares it. -2.
- src/Colors/RatedColor.php:L28: shrink: md5 over imploded channel values used only as an array dedup key - the imploded string already is the key. Drop the md5.

net: -250 lines possible.
```

src/Analyzers/DominantPaletteAnalyzer.php
olivervogel: Thanks. That's a good finding.

olivervogel:
@deluxetom Thanks for the findings.

src/Drivers/Gd/Analyzers/ColorCountAnalyzer.php
olivervogel: This seems to be not true.

I read a low color palette png image (color count is 4) turn it to truecolor and the color count is still 4.

```php
$image = imagecreatefrompng('image.png');
$state = imageistruecolor($image);
$totalOriginal = imagecolorstotal($image);
imagepalettetotruecolor($image);
$totalTruecolor = imagecolorstotal($image);

var_dump($state, $totalOriginal, $totalTruecolor);
```

```
bool(false)
int(4)
int(4)
```

src/Drivers/Imagick/Analyzers/ColorCountAnalyzer.php
olivervogel: It is only called once per color extraction. Do you really think this is a performance issue?

olivervogel:
> Review focused on the findings with speed impact or bug potential (terse one-line style, one per location). Lower-severity cleanups omitted.

I adressed almost all of the reportings. Also redesigned and refactored alot of code to allow for expandability of the swatches.

src/Analyzers/QuantizedPaletteAnalyzer.php
```diff
@@ -0,0 +1,78 @@
...
+    {
+        try {
+            $quantizer = $this->quantizer($image);
+        } catch (InvalidArgumentException $e) {
+            throw new AnalyzerException('Failed to analyze image colors', previous: $e);
+        }
+
+        return $quantizer->quantizeColors(
+            iterator_to_array($this->collectColors($image, $this->region)),
+        )->slice(0, $this->limit);
```
deluxetom: **`popular()` is never sorted by popularity.**

`slice(0, $this->limit)` keeps the *first* N bins in pixel-scan order. `Palette::addColor()` only appends, and nothing between it and here sorts by `Bin::count`. So the `PaletteInterface::popular()` contract ("sorted by popularity, highest first") and `$palette->first()` as "the most popular color" are both wrong.

Verified on `tests/resources/apple.jpg` with GD: `$image->colors()->popular(5)` returns bins with counts `775, 264, 93, 10, 50`. Descending order breaks at position 4, and bins with counts well above 10 are cut off entirely by the slice.

Suggested fix: add a `sortByCount()` to `Palette` (a `uasort` on `$this->bins` by `count` descending) and call it before slicing.

This is almost certainly why the `popular()` assertions in `tests/Unit/Colors/ColorExtractorTest.php:18-25` are commented out. They should be restored once the sort lands, since they are the only thing that would catch a regression here.

deluxetom (commented):
## Code review: Color Palette Extraction

Reviewed at PR head `d7064e99`. Every item below was reproduced by running PHP 8.5 with `gd` + `imagick` against the branch, not just read from the diff.

The architecture is good: `ColorExtractor` as an entry point, analyzers behind `AbstractPaletteAnalyzer`, `Quantizer`/`Bin`/`Palette` as the histogram layer, and swatches split into analyzer + filter. The problems are concentrated in `Palette` and in the quantization defaults.

> [!CAUTION]
> **`popular()` does not return the most popular colors.** `QuantizedPaletteAnalyzer` slices the first N bins in pixel-scan order and nothing ever sorts by `Bin::count`. On `apple.jpg` the top 5 come back with counts `775, 264, 93, 10, 50`. This contradicts the `PaletteInterface::popular()` docblock and is why the `popular()` assertions in `tests/Unit/Colors/ColorExtractorTest.php:18-25` are commented out.

### Correctness

| # | Location | Issue |
|---|---|---|
| 1 | `QuantizedPaletteAnalyzer.php:50` | `popular()` is unsorted, see above |
| 2 | `Palette.php:149` | `colorCount()` fatals on an unknown color instead of returning `0` |
| 3 | `Palette.php:169` | `toColorspace()` leaves array keys hashed from the *old* colorspace, so every later lookup misses and identical colors never merge |
| 4 | `Palette.php:210` | `sortByChannel()` reads the colorspace from `first()` only, so a mixed-colorspace palette throws |
| 5 | `Palette.php:64` | `offsetSet()` / `offsetUnset()` mutate a throwaway array, so `$palette[0] = $c` and `unset($palette[0])` are silent no-ops |
| 6 | `AbstractPaletteAnalyzer.php:44` | region guard uses `fitsWithin()`, which ignores the pivot, so an offset region samples outside the image and throws later |
| 7 | `AbstractSwatches.php:87` | `offsetUnset()` unsets a typed property, permanently breaking the object |
| 8 | `DominantPaletteAnalyzer.php:58` | the RNG is seeded once per instance, so "deterministic" only holds for the first `analyze()` call |
| 9 | `ColorExtractor.php:61` | user class is instantiated before the `SwatchesInterface` check, so an undeclared `ArgumentCountError` / `Error` escapes |

### Color quality

> [!WARNING]
> Every color the palette reports is a **bin center**, not a color that exists in the image. At the default level 20, solid black comes back as `rgb(6 6 6)` and `rgb(200 30 40)` as `rgb(198 32 45)`. `Bin` already carries a `color` field, so it could hold the first (or mean) real color of the bin and use the quantized triple only as the key.

- **Driver divergence** (`Drivers/Gd/Analyzers/ColorCountAnalyzer.php:21`): `imagecolorstotal()` returns `0` for truecolor images, so the "<= 256 colors, use `LEVEL_MAX`" fast path fires on Imagick and never on GD. A 4-color truecolor PNG gives exact colors on Imagick and quantized ones on GD.
- **Alpha fragments the histogram** (`Quantizer.php:58`): alpha stays unquantized while RGB is binned, so `popular(8)` on `tests/resources/300dpi.png` returns `rgb(45 236 249 / 0.82)` and `rgb(45 236 249 / 0.69)` as two separate top entries.
- **Swatch population is dead weight** (`VibrantMuted.php:37`): asking for a `256 * 256 * 256` palette selects level 256, so 2400 samples produce 996 unique colors and the largest population ratio is `0.0179`. Against saturation and lightness terms up to `9.0`, `WEIGHT_POPULATION` contributes nothing and swatch choice is decided purely by HSL distance.

### Performance

`DominantPaletteAnalyzer::updateCentroids()` rescans the full assignment list once per cluster, so k-means is `O(k*n)` per iteration across 50 iterations, with 3 to 6 accessor calls per distance. Measured on GD:

| Input | Call | Time |
|---|---|---|
| 2000x1500 | `dominant(8)` | 0.65s |
| 2000x1500 | `dominant(16)` | 3.05s |
| 2000x1500 | `dominant(64)` | 16.54s |
| 6000x4500 | `dominant(8)` | 9.07s |

Flattening colors to float triples once and accumulating per-cluster sums in a single pass over `$assignments` makes it `O(n)`.

### API surface

`Image::colors()` is only on the concrete class, not on `ImageInterface`. Since every public entry point is typed `: ImageInterface`, `$manager->read($path)->colors()` is a static-analysis error. `colorAt()`, `colorsAt()` and `reduceColors()` are all declared on the interface. The new tests pass only because the test helpers are typed against the concrete `Image`.

### CI and housekeeping

> [!IMPORTANT]
> `Palette.php:269` carries an unused `@phpstan-ignore missingType.checkedException`. PHPStan 2.1.1 at the project's level 6 reports this as an error ("No error with identifier missingType.checkedException is reported on line 270"), so static analysis will fail.

<details>
<summary>Minor cleanups</summary>

- `Quantizer.php:96` clamps a value that line 92 already guarantees is in range, and `binIndexToNormalized()` declares an `@throws RuntimeException` it can never throw.
- `AbstractSwatches::getIterator()` and `swatchNames()` run the same reflection filter twice, each starting with a dead `$x = []` assignment.
- `Palette` re-implements `first()`, `last()`, `count()`, `slice()`, `toArray()` and `getIterator()`, all of which `src/Collection.php` already provides.
- The blank-line removal in `src/Collection.php:187` is unrelated diff noise.
- Several new test files order `use` statements non-alphabetically (`tests/Unit/Analyzers/DominantPaletteAnalyzerTest.php`, `tests/Unit/Colors/Swatches/AbstractSwatchesTest.php`), contrary to `AGENTS.md` 4.3.
- `tests/Unit/AbstractPaletteAnalyzerTest.php` sits in `tests/Unit/` rather than `tests/Unit/Analyzers/` and has no `#[CoversClass]` attribute.

</details>

src/Colors/Palette.php
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @see PaletteInterface::colorCount()
+     */
+    public function colorCount(ColorInterface $color): int
+    {
+        return $this->bins[$this->hashColor($color)]->count;
```
deluxetom: **Fatals instead of returning `0` when the color is not in the palette.**

An unknown hash indexes `$this->bins` with a missing key and then reads `->count` on `null`:

```php
(new Palette([Color::rgb(255, 0, 0)]))->colorCount(Color::rgb(9, 9, 9));
// Warning: Undefined array key "..."
// Warning: Attempt to read property "count" on null
// TypeError: colorCount(): Return value must be of type int, null returned
```

`VibrantMutedFilter::findBestColor()` calls this, so it is reachable from `swatches()`.

```suggestion
        return $this->bins[$this->hashColor($color)]->count ?? 0;
```

src/Colors/Palette.php
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @see PaletteInterface::toColorspace()
+     */
+    public function toColorspace(string|ColorspaceInterface $colorspace): PaletteInterface
+    {
+        array_walk($this->bins, function (Bin $bin) use ($colorspace): void {
```
deluxetom: **`toColorspace()` leaves the array keys hashed from the old colorspace.**

`array_walk` rewrites `Bin::color` in place but the keys of `$this->bins` still encode the *previous* colorspace. Two consequences:

1. Every hash-based lookup afterwards misses.
2. Colors that become identical after conversion are never merged, so counts stay split.

```php
$p = new Palette([Color::rgb(255, 0, 0)]);
$p->toColorspace(Cmyk::class);
foreach ($p as $c) {
    $p->colorCount($c); // TypeError: Return value must be of type int, null returned
}
```

This also breaks `VibrantMutedFilter::findBestColor()` for any palette a caller converted first.

Suggested fix: rebuild `$this->bins` with re-hashed keys instead of walking in place, merging `count` when two converted colors collide on the same hash.

src/Colors/Palette.php
```diff
@@ -0,0 +1,318 @@
...
+
+        // normalize to channel classname
+        $channel = is_string($channel) ? $channel : $channel::class;
+        if (!class_exists($channel)) {
+            throw new InvalidArgumentException(
+                'The channel class "' . $channel . '" does not exist',
+            );
+        }
+
+        $originalColorspace = $this->first()->colorspace()::class;
```
deluxetom: **Mixed-colorspace palettes throw here.**

The "original colorspace" is derived from `first()` alone, and the conversion below only runs when that one differs from `$sortColorspace`. Colors in a *different* space than `first()` are never converted:

```php
(new Palette([Color::rgb(255, 0, 0), Color::cmyk(0, 100, 100, 0)]))
    ->sortByChannel(Red::class);
// InvalidArgumentException: Color channel ...Rgb\Channels\Red could not be found
```

`addColor()` accepts any `ColorInterface`, so mixed palettes are reachable through the public API. Suggested fix: decide per color, converting each one whose own `colorspace()` differs from `$sortColorspace`, rather than deciding once from `first()`.

src/Colors/Palette.php
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @see ArrayAccess::offsetSet()
+     */
+    public function offsetSet(mixed $offset, mixed $value): void
+    {
+        $this->toArray()[$offset] = $value;
```
deluxetom: **`offsetSet()` and `offsetUnset()` are silent no-ops.**

`toArray()` builds a fresh array from `$this->bins`, so both methods mutate a temporary that is discarded immediately:

```php
$p[0] = Color::rgb(1, 2, 3);
$p->first();  // still rgb(255 0 0)
unset($p[0]);
count($p);    // still 2
```

Writes vanish with no error, which is worse than an exception. Two options:

1. Implement both against `$this->bins` (set: re-hash and replace the bin at that index; unset: drop the bin).
2. Make the read-only intent explicit and throw `BadMethodCallException` from both.

Option 2 is probably the honest one for a histogram, since `$palette[0] = $color` has no well-defined count semantics.

Separately, `offsetExists()` and `offsetGet()` rebuild the whole color array on every single access.

src/Colors/Palette.php
```diff
@@ -0,0 +1,318 @@
...
+                    $colorA = $colorA->toColorspace($sortColorspace);
+                    $colorB = $colorB->toColorspace($sortColorspace);
+                }
+
+                return $colorA->channel($channel)->value() <=> $colorB->channel($channel)->value();
+            },
+        );
+
+        // reorder colors based on sorted indices and preserve keys
+        // @phpstan-ignore missingType.checkedException
```
deluxetom, line 374: **This unused ignore will fail CI.**

PHPStan 2.1.1 at the project's level 6 reports:

```
No error with identifier missingType.checkedException is reported on line 270
```

`array_combine()` throws nothing checked here, so the ignore has no target.

```suggestion
```

src/Analyzers/AbstractPaletteAnalyzer.php
```diff
@@ -0,0 +1,71 @@
...
+     * Get dynamic grid of pixel sample coordinates according to current image size.
+     *
+     * @throws InvalidArgumentException
+     * @return Generator<array{x: int, y: int}>
+     */
+    protected function sampleCoordinates(SizeInterface $size, ?SizeInterface $region = null): Generator
+    {
+        $region = $region === null ? $size : $region;
+
+        if (!$region->fitsWithin($size)) {
```
deluxetom: **`fitsWithin()` ignores the pivot, so an offset region passes and then samples out of bounds.**

`fitsWithin()` compares width and height only. A small region positioned near the edge passes this guard and then blows up further down in `PixelColorAnalyzer`:

```php
// 300x200 image
$image->colors()->popular(4, (new Size(10, 10))->setPivot(new Point(295, 195)));
// passes the guard (10 <= 300, 10 <= 200), then:
// InvalidArgumentException: The specified position (300, 195) is not within the image area
```

The error the user sees is also misleading, since it points at a pixel coordinate rather than at the region they passed.

Suggested fix: validate the region including its origin, for example `pivot()->x() + width() <= $size->width()` and the same for `y`, or clamp `$endX` / `$endY` to the image bounds.

src/Drivers/Gd/Analyzers/ColorCountAnalyzer.php
```diff
@@ -0,0 +1,23 @@
...
+
+        return $count === 0 ? null : $count;
```
deluxetom: **This makes GD and Imagick return different palettes for the same file.**

`imagecolorstotal()` returns `0` for truecolor GD images, which is mapped to `null` here. `QuantizedPaletteAnalyzer`'s "<= 256 colors, use `LEVEL_MAX`" fast path therefore fires on Imagick and never on GD.

Verified with a 24-bit truecolor PNG holding exactly 4 colors (`200,30,40` / `30,200,40` / `30,40,200` / `10,10,10`):

| Driver | `colorCount()` | `popular(8)` |
|---|---|---|
| GD | `null` | `rgb(198 32 45)`, `rgb(32 198 45)`, `rgb(32 45 198)`, `rgb(6 6 6)` |
| Imagick | `4` | `rgb(200 30 40)`, `rgb(30 200 40)`, `rgb(30 40 200)`, `rgb(10 10 10)` |

Either GD needs a real distinct-color count for truecolor images (a pixel scan building a hash set, which is what the analyzer is about to do anyway), or the `LEVEL_MAX` branch must not hinge on a value only one driver can supply.

src/Colors/Quantizer.php
```diff
@@ -0,0 +1,114 @@
...
+     * Convert a bin index to the center value of that bin.
+     *
+     * @throws RuntimeException
+     */
+    private function binIndexToNormalized(int $bin): float
+    {
+        $bin = max(0, min($this->levels - 1, $bin));
+
+        // @phpstan-ignore missingType.checkedException
+        return ($bin + 0.5) / $this->levels;
```
deluxetom: **Dequantizing to the bin center means no reported color exists in the source image.**

At the default level 20 used for `limit <= 256`, solid black `(0,0,0)` comes back as `rgb(6 6 6)` and `(200,30,40)` as `rgb(198 32 45)`. Pure black and pure white can never be returned. A user asking "what is the dominant color of this logo?" gets `#C6202D` instead of `#C81E28`.

`Bin` already carries a `color` field. Suggested fix: store the first (or the running mean of) the *actual* colors assigned to the bin as its representative, and use the quantized triple only as the bin key. That keeps the histogram grouping while returning colors that really occur in the image.

src/Colors/Quantizer.php
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * Return a quantized version of the given color.
+     *
+     * @throws ColorException
+     */
+    public function quantizeColor(ColorInterface $color): ColorInterface
+    {
+        try {
+            // preserve alpha unquantized
+            $alpha = $color->alpha()->normalized();
```
deluxetom: **Unquantized alpha fragments the histogram.**

RGB is binned but alpha is not, so visually identical colors that differ only slightly in alpha land in separate bins and compete against each other for the top slots.

`popular(8)` on `tests/resources/300dpi.png` returns `rgb(45 236 249 / 0.82)` and `rgb(45 236 249 / 0.69)` as two of the top entries. The same color takes two slots, pushing genuinely different colors out of the result.

Suggested fix: quantize alpha too, with its own (coarser) level, or leave alpha out of the bin key entirely and carry it on the representative color.

src/Image.php
```diff
@@ -1210,6 +1211,14 @@ public function encodeUsingPath(string $path, mixed ...$options,): EncodedImageI
         return $this->encode(new FilePathEncoder($path, ...$options));
     }
 
+    /**
+     * Get color extractor for the current image.
+     */
+    public function colors(): ColorExtractor
```
deluxetom, line 1217: **`colors()` is missing from `ImageInterface`, so the feature is unreachable from the type every entry point returns.**

```php
(new ImageManager(Driver::class))->decodePath($p)->colors();
// decodePath() is typed : ImageInterface, which has no colors()
// PHPStan / IDE: Call to an undefined method ImageInterface::colors()
```

Every comparable feature (`colorAt()`, `colorsAt()`, `reduceColors()`) is declared on the interface. The new tests only pass because the test helpers are typed against the concrete `Image` class.

While adding it, consider returning an interface rather than the concrete `ColorExtractor`, to match how `colorsAt()` returns `CollectionInterface`.

src/Analyzers/DominantPaletteAnalyzer.php
```diff
@@ -0,0 +1,329 @@
...
+     * @throws InvalidArgumentException
+     */
+    public function __construct(protected int $limit = 8, protected ?SizeInterface $region = null)
+    {
+        if ($this->limit < 1) {
+            throw new InvalidArgumentException('Invalid $limit value. Must be int<1, max>');
+        }
+
+        // use local RNG to avoid corrupting global mt_rand state
+        $this->rng = new Randomizer(new Mt19937(self::SEED));
```
deluxetom: **Seeding in the constructor only makes the *first* `analyze()` deterministic.**

The `Randomizer` keeps its state between calls, so a reused analyzer gives different results for the same image:

```php
$a = new DominantPaletteAnalyzer(4);
$a->analyze($img); // rgb(95 196 205) | rgb(223 163 144) | rgb(163 47 42) | rgb(93 122 67)
$a->analyze($img); // rgb(95 196 206) | ...            | rgb(163 46 42) | ...
```

Suggested fix: re-seed at the top of `analyze()` so the output depends only on the image:

```php
$this->rng = new Randomizer(new Mt19937(self::SEED));
```

A regression test asserting that two consecutive `analyze()` calls on the same image are identical would lock this down.

src/Analyzers/DominantPaletteAnalyzer.php
```diff
@@ -0,0 +1,329 @@
...
+     * @return array<OklabColor>
+     */
+    private function updateCentroids(array $colors, array $assignments, int $k): array
+    {
+        $centroids = [];
+
+        for ($i = 0; $i < $k; $i++) {
+            $clusterColors = [];
+
+            foreach ($assignments as $colorIndex => $cluster) {
```
deluxetom: **This inner loop rescans every assignment once per cluster, making `dominant()` unusable on real images.**

`updateCentroids()` is `O(k*n)` per iteration across 50 iterations, and each distance and mean goes through 3 to 6 object accessor calls. Measured on GD:

| Input | Call | Time |
|---|---|---|
| 2000x1500 | `dominant(8)` | 0.65s |
| 2000x1500 | `dominant(16)` | 3.05s |
| 2000x1500 | `dominant(64)` | 16.54s |
| 6000x4500 | `dominant(8)` | 9.07s |

Two changes bring this down a long way:

1. Invert the loop, accumulating per-cluster sums and counts in a single pass over `$assignments` (`O(n)` instead of `O(k*n)`).
2. Flatten the colors to plain float triples once up front and compare on raw floats, instead of calling `->lightness()->value()` and friends inside the innermost loop.

src/Colors/ColorExtractor.php
```diff
@@ -0,0 +1,74 @@
...
+     */
+    public function swatches(string|SwatchesInterface $swatches = new Swatches\VibrantMuted()): SwatchesInterface
+    {
+        if (is_string($swatches) && !class_exists($swatches)) {
+            throw new InvalidArgumentException(
+                'The specified swatches class "' . $swatches . '" does not exist',
+            );
+        }
+
+        $swatches = is_string($swatches) ? new $swatches() : $swatches;
```
deluxetom: **The class is instantiated before it is checked, so the wrong exception escapes.**

If the named class exists but is abstract or has a required constructor argument, `new $swatches()` throws before line 64 ever runs, so callers get an undeclared `ArgumentCountError` or `Error: Cannot instantiate abstract class` instead of the intended `InvalidArgumentException`.

Checking the type before constructing fixes it and removes the need for the `instanceof.alwaysTrue` ignore:

```php
if (is_string($swatches)) {
    if (!is_a($swatches, SwatchesInterface::class, true)) {
        throw new InvalidArgumentException(
            'The specified swatches argument must be a classname of or implement ' . SwatchesInterface::class,
        );
    }

    $swatches = new $swatches();
}
```

Two related notes:

- The `@return T` generic is unsound. The returned object comes from `colorFilter()->filterColors()`, not from `$swatches`, so a custom `MySwatches` whose `colorFilter()` returns `VibrantMutedFilter` is documented as returning `MySwatches` but actually returns `VibrantMuted`.
- A passed *instance* is used only as an analyzer/filter factory and then discarded, so any pre-configured state on it is silently dropped. If instances are not meant to carry state, accepting only a class-string would be clearer.

src/Colors/Swatches/AbstractSwatches.php
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @see ArrayAccess::offsetUnset()
+     */
+    public function offsetUnset(mixed $offset): void
+    {
+        unset($this->{$offset});
```
deluxetom: **Unsetting a typed property leaves the object permanently broken.**

After `unset()`, the property is uninitialized rather than `null`, so every later read fatals:

```php
$sw = new VibrantMuted(vibrant: Color::rgb(1, 2, 3));
unset($sw['vibrant']);
$sw->vibrant;    // Error: Typed property VibrantMuted::$vibrant
$count = count($sw);  //   must not be accessed before initialization
$sw->toArray();
```

All swatch properties are declared `?ColorInterface`, so assigning `null` is both valid and the semantics callers expect from `unset()` on a swatch slot.

```suggestion
        $this->{$offset} = null;
```

src/Colors/Swatches/VibrantMuted.php
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * {@inheritdoc}
+     *
+     * @see SwatchesInterface::colorAnalyzer()
+     */
+    public function colorAnalyzer(): AnalyzerInterface
+    {
+        // @phpstan-ignore missingType.checkedException
+        return new QuantizedPaletteAnalyzer(256 * 256 * 256);
```
deluxetom: **Asking for a 16.7M color palette disables quantization and makes the population term dead weight.**

`256 * 256 * 256` selects `Quantizer` level 256, which is effectively no quantization, so nearly every sampled pixel becomes its own bin.

Measured on `apple.jpg`: the swatch palette holds **996 unique colors for 2400 samples**, and the largest population ratio is `0.0179`. So `populationRatio * WEIGHT_POPULATION` (1.0) contributes at most `0.018` against saturation and lightness terms reaching `9.0`. Swatch selection is decided purely by HSL distance, and the most frequent color in the image has effectively no influence on the result.

The reference Vibrant / Palette algorithms bin to roughly 16 to 32 levels per channel precisely so populations stay meaningful. Suggested fix: quantize to a coarse level here, which also removes the magic number and the need for the `missingType.checkedException` ignore above.

olivervogel:
Oh my, even more! Thanks anyway. 😂

deluxetom:
@olivervogel no problem, want me to post a PR to this branch with the fixes? I have some extra claude usage I can dedicate to this :)

olivervogel:
@deluxetom That would be really great. I pushed my last local commits.

olivervogel:
@deluxetom I've moved a few things around. Mainly put the quantization logic directly to the palette and added a method to quantize but keep original channel values.

Thanks again for the support.

deluxetom:
@olivervogel last review, almost there :) 
```
Worth fixing before merge (both cheap, both hit the documented $swatches['...'] API):

1. src/Colors/Swatches/AbstractSwatches.php:55 - offsetExists() uses property_exists(), so isset($swatches['vibrant']) returns true when the swatch is null (the normal "no match" result, see testFilterColorsNotFound). if (isset($swatches['lightMuted'])) { $swatches['lightMuted']->toHex(); } fatals. Use $this->{$offset} !== null.
2. AbstractSwatches.php:75 and :85 - offsetSet()/offsetUnset() accept any offset, so $swatches['foo'] = $color creates a dynamic property. Deprecated on PHP 8.3+, an Error in PHP 9. Guard against unknown offsets.

Worth a maintainer's opinion:

3. src/Analyzers/AbstractPaletteAnalyzer.php:25 - every sampled pixel calls $image->colorAt(), which goes through Image::analyze() and a full AbstractDriver::specialize(). Up to ~111k specializations per analysis on a 10000x10000 image. Reuse one specialized analyzer, or do a driver-level bulk pixel read.
```

olivervogel:
## Unfinished Business

~~### 1. Optimization of sample pixel collection~~

~~src/Analyzers/AbstractPaletteAnalyzer.php:25 - every sampled pixel calls $image->colorAt(), which goes through Image::analyze() and a full AbstractDriver::specialize(). Up to ~111k specializations per analysis on a 10000x10000 image. Reuse one specialized analyzer, or do a driver-level bulk pixel read.~~

✅ Done

~~### 2. Standardize SwatchesInterface and PaletteInterface~~

~~We could consider extending SwatchesInterface from PaletteInterface to standardize the handling, since they both do almost the same thing.~~

~~The main difference is that Palette objects use integers as iteration keys, while Swatches objects use strings.~~

~~The implementation of sorting functions in Swatches is also still undecided.~~

Swatches and palettes stay separated. If more functionality is needed it is always possible to convert to a palette. Renamed "swatches" to "theme".

~~### 3. Palettes and ColorFilters~~

~~I think it makes more sense overall if `ColorFilterInterface::filterColors()` returns its input (Palette objects) in the same type. Currently the method returns `SwatchesInterface`.~~

~~A new PaletteInterface::filter() method is a good option; it accepts `ColorFilterInterface` objects and returns `PaletteInterface` objects.~~

~~It's unclear where to put the swatch object instantiation.~~

Implemented `PaletteInterface::filter()` with simple callback. "Color filters" are now Palette definitions.

~~### 4. Pick most used color in reduce()~~

~~Pick most used color instead of just the first one.~~

✅ Done

PR #1510, opened 2026-08-14
Color palette extraction: apply remaining review fixes

deluxetom:
Follow-up to the review on #1509, as discussed. This applies the remaining findings on top of your latest commits. Each fix comes with a regression test.

## Correctness

- **`Palette::toColorspace()`** rebuilds the bins with re-hashed keys, so lookups work after conversion and colors that become identical are merged into one bin.
- **`Palette::sortByChannel()`** converts each color individually, so palettes with mixed colorspaces no longer throw.
- **`Palette::offsetSet()` / `offsetUnset()`** throw `NotSupportedException` instead of silently mutating a temporary array.
- **`AbstractPaletteAnalyzer`** validates the sample region including its position, so offset regions fail with a clear message instead of a misleading pixel coordinate error later.
- **`AbstractSwatches::offsetUnset()`** assigns `null` instead of `unset()`, which left the typed property uninitialized and made every later read fatal.
- **`DominantPaletteAnalyzer`** re-seeds its RNG on every `analyze()` call, so a reused instance returns identical results for the same image.
- **`ColorExtractor::swatches()`** validates the classname before instantiation, so abstract classes or constructors with required parameters produce the intended `InvalidArgumentException`. The unsound `@return T` generic is removed.

## Color quality

- **Bins keep the first actual color as representative** instead of the calculated bin center, so every reported color exists in the source image (solid black comes back as `rgb(0 0 0)`, not `rgb(6 6 6)`).
- **Alpha is left out of the bin key**, so colors that only differ slightly in transparency no longer occupy multiple palette slots.
- **`popular()` sorts by presence before applying the limit**, so the result contains the globally most frequent colors. The previously commented-out value assertions are restored and pass.
- **The `LEVEL_MAX` decision now counts distinct sampled colors** instead of using a driver-specific color count. GD and Imagick return identical palettes for the same file, and the full-resolution Imagick histogram pre-scan is gone as a side effect. The exact-color path for low-color images is kept.
- **`VibrantMuted` analyzes a coarsely quantized palette** (256 instead of 256³), so the population weight has an actual influence on swatch scoring.

## Performance

`DominantPaletteAnalyzer` flattens colors once to plain Oklab float triples and accumulates per-cluster sums in a single pass over the assignments, O(n) instead of O(k*n) per iteration. Measured on a 2000x1500 image (GD):

| Call | Before | After |
|---|---|---|
| `dominant(8)` | 0.64s | 0.04s |
| `dominant(16)` | 3.05s | 0.10s |
| `dominant(64)` | 16.43s | 0.43s |

## API

- `colors()` is declared in `ImageInterface`, so the feature is reachable from the type every entry point returns.

## Housekeeping

- The unrelated whitespace change in `src/Collection.php` is reverted.
- `AbstractPaletteAnalyzerTest` moves to `tests/Unit/Analyzers/` and receives a `CoversClass` attribute; `use` statements in the new test files are ordered alphabetically.

> [!NOTE]
> The three `ColorCountAnalyzer` classes (generic, GD, Imagick) are no longer used by the extractor. I kept them because they are public analyzers, but they can be dropped if you prefer.

<details>
<summary>Verification</summary>

- Full PHPUnit suite passes with GD and Imagick (PHP 8.5).
- PHPStan level 6: no errors.
- PHP CodeSniffer: no violations.
- GD and Imagick return byte-identical results for `popular()`, `dominant()`, and `swatches()` on the test images.

</details>

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Analyzers/AbstractPaletteAnalyzer.php
+++ b/src/Analyzers/AbstractPaletteAnalyzer.php
@@ -41,14 +41,21 @@ abstract class AbstractPaletteAnalyzer implements AnalyzerInterface
     {
         $region = $region === null ? $size : $region;
 
-        if (!$region->fitsWithin($size)) {
-            throw new InvalidArgumentException('The region must not be larger than the actual image size');
-        }
-
         $startX = $region->pivot()->x();
         $startY = $region->pivot()->y();
         $width = $region->width();
         $height = $region->height();
+
+        // validate the region including its position, otherwise offset
+        // regions would sample coordinates outside of the image
+        if (
+            $startX < 0 || $startY < 0
+            || $startX + $width > $size->width()
+            || $startY + $height > $size->height()
+        ) {
+            throw new InvalidArgumentException('The region must fit within the actual image size');
+        }
+
         $totalPixels = $width * $height;
 
         $sampleRate = match (true) {
--- a/src/Analyzers/DominantPaletteAnalyzer.php
+++ b/src/Analyzers/DominantPaletteAnalyzer.php
@@ -9,7 +9,6 @@ use Intervention\Image\Colors\Oklab\Colorspace as Oklab;
 use Intervention\Image\Colors\Palette;
 use Intervention\Image\Exceptions\AnalyzerException;
 use Intervention\Image\Exceptions\InvalidArgumentException;
-use Intervention\Image\Interfaces\ColorInterface;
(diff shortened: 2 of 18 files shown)


olivervogel:
@deluxetom Very nice. I'll finish it all up tomorrow.

PR #1511, opened 2026-08-17
Fix (gd): preserve alpha in TrimModifier reference color

cheack:
## Bug

`Image::trim()` on the GD driver silently no-ops when the image has a transparent margin.

## Root cause

`TrimModifier::trimColor()` averages the four corner pixels' RGB but allocates the `imagecropauto()` reference color with `imagecolorallocate()`, which always forces full opacity. When the real corner pixels are transparent, the reference color's alpha never matches them, so `imagecropauto()` finds nothing within tolerance and crops nothing.

## Fix

- Average and preserve alpha across the four corners.
- Allocate the reference color with `imagecolorallocatealpha()` instead of `imagecolorallocate()`.
- Set `imagealphablending(false)` / `imagesavealpha(true)` before sampling, so `imagecolorat()` returns raw unblended pixel values matching what gets allocated.


deluxetom (approved):

olivervogel:
Thank you. With the image you provided everything works as expected now.

I tested with this image and the transparent areas are replaced by black.

out

That was already the case before your patch. Unfortunately, I won't be able to take a look at this until the next few days. Feel free to check it out.

nlemoine:
I hit a related problem from the other side. I think the averaging itself is the root cause, not only the alpha part.

`trim()` on GD also does nothing when the four corners are opaque but disagree. This can be reproduced without any transparency, with the 120x120 PNG attached below. Three corners white, bottom-right red.

trim-corners

```
->trim()   gd        120x120   (nothing trimmed)
           imagick    70x70
```

Same cause as yours, I think. `trimColor()` averages the corners into `rgb(255, 191, 191)`, and the image contains that color nowhere, so `imagecropauto()` matches nothing. Averaging the alpha has the same failure mode, three transparent corners plus one opaque one average to something that matches neither.

I tried dropping the averaging and using the raw top-left pixel instead:

```php
private function trimColor(ImageInterface $image): int
{
    return imagecolorat($image->core()->native(), 0, 0);
}
```

`imagecolorat()` already carries the alpha on truecolor images, so it covers the transparent case as well. Measured on `develop` (`0a5aa57a`), PHP 8.3.33:

```
image                            imagick   gd today   gd top-left
trim.png, tol 0                  28x28     28x28      28x28
radial.png, tol 50               23x23     35x35      35x35
trim-transparent.png, tol 0      25x25     41x47      25x25
120x120 3 white corners, tol 0   70x70     120x120    70x70
uniform 50px border, tol 0       100x100   100x100    100x100
```

Your fixture lands on 25x25 too, and none of the existing expectations move. Full suite unchanged here, same 21 failures before and after, all from my local ImageMagick build missing JPEG2000 and OKLAB.

Not full parity though. With a white border on top and left and a black one on right and bottom, imagick gives 150x150 and the top-left version still gives 200x200. Imagick seems to use a different corner per edge, which `imagecropauto()` cannot do in one pass.

No idea about the black transparent areas @olivervogel mentioned, I did not look at that.

Worth widening it here, or should I open a separate one for the opaque case?

olivervogel:
> `trim()` on GD also does nothing when the four corners are opaque but disagree. This can be reproduced without any transparency, with the 120x120 PNG attached below. Three corners white, bottom-right red.

@nlemoine I think that using this average color will always lead to edge cases. Not a very reliable strategy.

olivervogel:
I have addressed this issue in a broader context, including the black background (which was not fixed here) in PR #1512. Please create a new issue if something is missing.

PR #1517, opened 2026-08-22
Apply Imagick colorspace change to every frame

nlemoine:
`Imagick::transformImageColorspace()` only mutates the frame the iterator is parked on, so `setColorspace()` on an animated image leaves the other frames alone.

Repro, a 2 frame TIFF where both frames are CMYK:

```
decoded by Intervention:   f0=CMYK f1=CMYK
after setColorspace(Rgb):  f0=sRGB f1=CMYK
```

Which frame gets converted depends on where the iterator sits, not on the image. Call `colorsAt()` first and you get `f0=CMYK f1=sRGB` instead.

It gets worse downstream, `InsertModifier` does iterate every frame, so a half converted base composites an sRGB watermark into CMYK bands on the frames the change missed.

The fix iterates, same shape `GrayscaleModifier` already uses for the identical call. GD is a no-op in this path.

Four other modifiers had the same shape, so they are in here too. Measured on the 8 frame `animation.gif`:

| modifier | call | before |
| --- | --- | --- |
| `ResolutionModifier` | `setImageResolution()` | x-resolution `300,0,0,0,0,0,0,0` |
| `ProfileModifier` | `profileImage()` | profile on `yes,no,no,no,no,no,no,no` |
| `RemoveProfileModifier` | `profileImage(null)` | cleared on one frame only |
| `StripMetaModifier` | `stripImage()` | comment survives on every frame but one |

`StripMetaModifier` is the one I would worry about. Every encoder runs it when strip is enabled, and an animated TIFF written with `strip: true` kept the meta data of all its frames but one. Webp and avif are not affected, those writers do not round-trip per frame metadata at all. Its `setExif()` reset stays outside the loop, that one is a property of the image and not of a frame.

`TrimModifier` has the same shape but refuses animated images already, and the rest delegate to modifiers that iterate.

Every fix has an animated test, all of them fail on `develop`. There was no Imagick specific test for `ColorspaceModifier`, `ProfileModifier` or `RemoveProfileModifier` at all before this.

Two other places have the same fault and are not in here, `NativeObjectDecoder` and `ReduceColorsModifier`. Details in the comments.

Note if you write your own repro: building the fixture with `addImage()` twice and then one `transformImageColorspace(CMYK)` gives you `f0=CMYK f1=sRGB`, for the same reason. Each frame has to be converted before it is added.

### DIFF
--- a/src/Drivers/Imagick/Modifiers/ColorspaceModifier.php
+++ b/src/Drivers/Imagick/Modifiers/ColorspaceModifier.php
@@ -28,24 +28,27 @@ class ColorspaceModifier extends GenericColorspaceModifier implements Specialize
      */
     public function apply(ImageInterface $image): ImageInterface
     {
-        $colorspace = $this->targetColorspace();
-        $imagick = $image->core()->native();
+        $target = $this->imagickColorspaceOrFail($this->targetColorspace());
 
-        try {
-            $result = $imagick->transformImageColorspace(
-                $this->imagickColorspaceOrFail($colorspace),
-            );
+        // Imagick::transformImageColorspace() only affects the frame the
+        // iterator is currently pointing at, so every frame has to be
+        // transformed individually. Otherwise all frames but one would keep
+        // their original colorspace on animated images.
+        foreach ($image as $frame) {
+            try {
+                $result = $frame->native()->transformImageColorspace($target);
 
-            if ($result === false) {
+                if ($result === false) {
+                    throw new ModifierException(
+                        'Failed to apply ' . self::class . ', unable to transform image colorspace',
+                    );
+                }
+            } catch (ImagickException $e) {
                 throw new ModifierException(
                     'Failed to apply ' . self::class . ', unable to transform image colorspace',
+                    previous: $e,
                 );
             }
-        } catch (ImagickException $e) {
-            throw new ModifierException(
-                'Failed to apply ' . self::class . ', unable to transform image colorspace',
-                previous: $e,
-            );
         }
 
         return $image;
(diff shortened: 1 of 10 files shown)


olivervogel:
Thank you @nlemoine. If this issue really does apply to other modifiers as well, feel free to add a fix to this PR.

nlemoine:
It does, four of them. Measured on the 8 frame `animation.gif`, all fixed here:

| modifier | before |
| --- | --- |
| `ResolutionModifier` | x-resolution `300,0,0,0,0,0,0,0` |
| `ProfileModifier` | profile on `yes,no,no,no,no,no,no,no` |
| `RemoveProfileModifier` | clears one frame |
| `StripMetaModifier` | comment survives on every frame but one |

`StripMetaModifier` is the one worth looking at, every encoder runs it when strip is enabled. An animated TIFF written with `strip: true` kept the meta data of every frame but one. I had written that webp and avif were affected too, then checked and they are not, those writers do not round-trip per frame metadata at all.

`TrimModifier` has the same shape but refuses animated images already. The rest delegate to modifiers that iterate. Each fix has a test that fails on `develop`.

Two more places have it, not in this PR, tell me if you want them.

`NativeObjectDecoder`, its GRAY to sRGB normalisation only reaches the first frame. A 3 page grayscale tiff decodes to colorspaces `23,3,3`. Inserting a red watermark on every frame then gives `255,0,0 | 255,255,255 | 255,255,255`, which is the exact thing the comment there says it is preventing.

`ReduceColorsModifier`, its early return reads one frame's color count:

```php
$image->reduceColors(4);            // colors 8,8,1 -> 3,3,1   ok
$image->blur(1)->reduceColors(4);   // colors 8,8,1 -> 8,8,1   skipped
```

Any iterating modifier moves the pointer off frame 0 and a single low color frame then skips the whole animation. Careful with this one, moving the check into the loop breaks `PngEncoderTest::testEncoderIndexed`, just entering the frame loop changes the output. Not a one liner, which is why I left it out.

The PR title is narrower than what is in it now, rename it if you prefer.

olivervogel:
Thank you Nicolas.

PR #1522, opened 2026-08-24
Remove JsonSerializable from EncodedImage

### LINKED ISSUE #1521: image `content-type` changed by Laravel
## Describe the bug

Hi,

After the `JsonSerializable` interface was added in #1513, the content-type of image responses is being altered by Laravel. 

Laravel automatically changes the content-type from `image/jpeg` to `application/json` when the response content is an instance of `JsonSerializable`. This behavior can be observed in the Laravel framework code:

https://github.com/laravel/framework/blob/d46a1cc4e8736600fe34f8607f55d79f44664e38/src/Illuminate/Http/Response.php#L90-L103

Additionally, the `mewebstudio/captcha` package, which relies on this library to generate captcha images, is affected by this issue:

https://github.com/mewebstudio/captcha/blob/b7add434bfd87c7e900c82d821a56ec566381915/src/Captcha.php#L340-L347

**Expected behavior:** The response should maintain the correct `image/jpeg` content-type when returning image data.

**Actual behavior:** The content-type is incorrectly set to `application/json`, causing image rendering to fail in browsers and API clients.

## Images

**Version 4.2 (Working correctly):**
Image

**Version 4.3 (Broken):**
Image

## Environment (please complete the following information):
- PHP Version: 8.4.17
- OS: Ubuntu
- Intervention Image Version: 4.3

lazerg:
JsonSerializable on EncodedImage (added in #1513) makes Laravel's response layer treat any encoded image as JSON content and rewrite the response Content-Type to application/json, breaking `return response($image->encode())`, the most common way this library is used in HTTP contexts (also affects mewebstudio/captcha).

Color, DataUri, and Resolution keep JsonSerializable since they aren't normally returned as raw HTTP bodies. EncodedImage is the one that regularly is, so this just drops the interface there.

Fixes #1521

### DIFF
--- a/src/EncodedImage.php
+++ b/src/EncodedImage.php
@@ -8,10 +8,9 @@ use Intervention\Image\Exceptions\InvalidArgumentException;
 use Intervention\Image\Exceptions\StreamException;
 use Intervention\Image\Interfaces\DataUriInterface;
 use Intervention\Image\Interfaces\EncodedImageInterface;
-use JsonSerializable;
 use Throwable;
 
-class EncodedImage extends File implements EncodedImageInterface, JsonSerializable
+class EncodedImage extends File implements EncodedImageInterface
 {
     /**
      * Create new instance.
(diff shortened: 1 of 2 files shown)


olivervogel:
Yeah, maybe it's better this way. Thanks.

PR #1529, opened 2026-09-21
Fix decoding of Unicode image file paths

Gruven:
Image paths containing non-ASCII characters, such as `фото.jpg` or `图片.jpg`, are incorrectly classified as binary data. The binary decoder fails before the file path decoder is reached.

This change gives existing files precedence over the binary-data heuristic and allows existing non-ASCII paths through file path detection. Empty, oversized and NUL-containing paths are rejected before filesystem checks.

Regression tests cover Cyrillic and Chinese filenames, absolute and relative paths, and string and Stringable inputs with both GD and Imagick. Decoder order remains unchanged.


olivervogel:
These methods only check whether the input COULD be the result. Hence the naming “couldBe*” they should not reach the file system.

I fixed this with an other approach in https://github.com/Intervention/image/pull/1530

Having said that, thanks for letting me know.

